PR #7, opened 2022-03-24
Change some result types

gustavoguichard:
This PR does a few things:
- Removes the `UnpackError` utility type as its result is always going to be the same as `ErrorResult` so we can just use that type instead
- Adds `UnpackResult` utility function, to complement the `UnpackSuccess` and `UnpackData` so we can extract all specific types out of domain functions
- Move types to a single file to simplify the exports of `index.ts`
- Updates `README.md`
- Adds `errors` and `inputErrors` as empty arrays on the `SuccessResult`. So now we can have a better DX when developing React components. See example below:

## Assuming the following Remix action:
```tsx
const myDomainFunction = makeDomainFunction()(async () => ({ name: 'John' }))

export const action = async () => {
  const result = await myDomainFunction()
  return json(result, { status: result.success ? 200 : 422 })
}
```

## Previously
If we were to return the whole result from an action, we'd have to make checks in the UI like so:
```tsx
export default Component() {
  const actionData = useActionData<UnpackResult<typeof myDomainFunction>>()
  return (
  <>
    <form>
      {/* ... */}
      {actionData?.success === false && actionData.errors.map(e => <span>{e.message}</span>)}
    </form>
  </>
  )
}
```

## Now
With the addition of empty errors to the success type and null data to the error type the component code will be simplified to:
```tsx
export default Component() {
  const actionData = useActionData<UnpackResult<typeof myDomainFunction>>()
  return (
  <>
    <form>
      {/* ... */}
      {actionData.errors.map(e => <span>{e.message}</span>)}
    </form>
  </>
  )
}
```

### DIFF
--- a/src/domain-functions.ts
+++ b/src/domain-functions.ts
@@ -29,14 +29,12 @@ const makeDomainFunction: MakeDomainFunction =
           success: false,
           errors: [],
           inputErrors: result.error.issues,
-          data: null,
         }
       } else if (envResult.success === false) {
         return {
           success: false,
           errors: envResult.error.issues,
           inputErrors: [],
-          data: null,
         }
       }
       try {
@@ -48,7 +46,7 @@ const makeDomainFunction: MakeDomainFunction =
         }
       } catch (error) {
         const errors = [{ message: (error as Error).message }]
-        return { success: false, errors, inputErrors: [], data: null }
+        return { success: false, errors, inputErrors: [] }
       }
     }) as DomainFunction<Awaited<ReturnType<typeof handler>>>
     return domainFunction
--- a/README.md
+++ b/README.md
@@ -40,7 +40,6 @@ failedResult = {
   success: false,
   inputErrors: [{ path: ['number'], message: 'Expected number, received nan' }],
   errors: [],
-  data: null,
 }
 */
 ```
@@ -130,7 +129,6 @@ type ErrorResult = {
   success: false
   errors: z.ZodIssue[] | { message: string }
   inputErrors: z.ZodIssue[]
-  data: null
 }
 ```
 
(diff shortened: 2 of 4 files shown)


src/domain-functions.ts
```diff
@@ -33,43 +25,33 @@ const makeDomainFunction: MakeDomainFunction =
...
         return {
           success: true,
           data: await handler(result.data, envResult.data),
+          errors: [],
+          inputErrors: [],
         }
       } catch (error) {
         const errors = [{ message: (error as Error).message }]
-        return { success: false, errors, inputErrors: [] }
+        return { success: false, errors, inputErrors: [], data: null }
```
diogob, line 51: I think I rather have `data` only in the success case. Here it seems to add more code without real benefit on the consumer side (since it was already `undefined` before).

diogob: What bothers me with the data null in the type is that the `.data` in the result will type check even when we know that is a useless value (at least in the case of the empty lists they are mapped over).
In the code you have used as motivation we still have the conditionals in the form of the safe navigation and coalesce operators:
```
    Submitted: {actionData.data?.name ?? ''}
```

I would rather have the typechecker tell me data does not exist until I narrow the type, so I have a branch where success is guaranteed. I could even argue that usually the code above would be something like:

```
    {actionData.success ? 'Submitted: ' + {actionData.data.name} : 'Failure to sumit'}
```

gustavoguichard: Ok, I think I didn't come up with the best example and the need might emerge but I'll agree with you for now and let it emerge if it ever will.

I'll rollback the nullish data for now

gustavoguichard:
@diogob LMK what you think now

diogob (approved):
Looks great

PR #13, opened 2022-04-12
Add mapError combinator

diogob:
## Purpose

Oftentimes is useful to transform errors produced in a domain function.
This PR introduces the `mapError` combinator, very similar to the existing `map` but operates on the error side of the `Result` type.

## Technical details

Just like [mapLeft](https://hackage.haskell.org/package/either-5.0.1.1/docs/Data-Either-Combinators.html#v:mapLeft) but for domain functions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ It does this by enforcing the parameters' types in runtime (through [zod](https:
   - [all](#all)
   - [pipe](#pipe)
   - [map](#map)
+  - [mapError](#mapError)
 - [Input Utilities](#input-utilities)
   - [inputFromForm](#inputfromform)
   - [inputFromUrl](#inputfromurl)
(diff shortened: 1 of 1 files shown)


src/domain-functions.ts
```diff
@@ -130,5 +130,27 @@ const map: Map = (dfn, mapper) => {
     }
   }
 }
+type MapError = <O>(
+  dfn: DomainFunction<O>,
+  mapper: (element: ErrorData) => ErrorData,
```
gustavoguichard, line 135: Looking at the implementation I don't understand why the tsc didn't complain about this:
```suggestion
  mapper: (element: ErrorResult) => ErrorData,
```

gustavoguichard, line 135: Looking at the implementation I don't understand why the tsc didn't complain about this:
```suggestion
  mapper: (element: ErrorResult) => ErrorData,
```

diogob: `ErrorData` fits in `ErrorResult`. For the mapper function `ErrorData` makes more sense since we don't wan't the developer to receive or return the `success` boolean.

gustavoguichard (approved):

PR #14, opened 2022-04-13
New error format

diogob:
## Purpose

This PR add some utilities to format errors with ease. It should facilitate its use in [remix-forms](https://github.com/SeasonedSoftware/remix-forms) but also create an overall more pleasant DX when handling error cases.

## Technical details

We dropped the old and undocumented `formatErrors` and included 2 more generic functions `errorsByName` and `errorsForSchema`. For more details check the README diff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -462,7 +462,7 @@ To better understand how to structure your data, refer to [qs documentation](htt
 To improve DX when dealing with errors we do export a couple of utilities.
 
 ### errorMessagesFor
-Given a array of `SchemaError` be it from `inputErrors` or `environmentErrors` and a name, it returns the first error with that name in its path.
+Given a array of `SchemaError` be it from `inputErrors` or `environmentErrors` and a name, it returns a list of error messages with that name in their path.
 
 ```tsx
 const result = {


README.md
```diff
@@ -442,6 +458,44 @@ export const action = async ({ request }) => {
 
 To better understand how to structure your data, refer to [qs documentation](https://github.com/ljharb/qs#parsing-objects)
 
+## Error Utilities
+To improve DX when dealing with errors we do export a couple of utilities.
+
+### errorsByName
```
danielweinmann, line 464: Shouldn't this be called `errorByName` in singular form?

danielweinmann: Or maybe `errorOn`?

danielweinmann: Hehe, actually, I think these helpers should be referring to `messages` and not `errors`, because we're returning messages and not the error objects.

Maybe `messagesFor` and `messagesForSchema`?

danielweinmann (changes requested):

README.md
gustavoguichard: I'd rather have both words to easily know what they refer to.. `errorMessagesFor`, `errorMessagesForSchema`

diogob: I settled with your suggestion @gustavoguichard

PR #17, opened 2022-04-19
environment type inference

diogob:
## Purpose

The environment parameter is usually fully described by the developer, as opposed to the input that can be grabbed from some object without shape.
Having the environment type-checked can help developers to catch some errors early on, such as a change in some authorization requirement (when a domain privilege is passed through the environment).

## Technical details

So far there are some problems with the type inference on compositions:

- All `map`, `mapError` environment types have their keys turned to optional.
- I have not figured out a way to correctly infer the type on the `all` and `pipe` combinators.

### DIFF
--- a/src/domain-function.test.ts
+++ b/src/domain-function.test.ts
@@ -105,6 +105,27 @@ describe('all', () => {
     })
   })
 
+  it('should use the same environment', async () => {
+    const a = makeDomainFunction(
+      z.undefined(),
+      z.object({ toIncrement: z.number() }),
+    )(async (_input, { toIncrement }) => toIncrement + 1)
+    const b = makeDomainFunction(
+      z.undefined(),
+      z.object({ toDecrement: z.number() }),
+    )(async (_input, { toDecrement }) => toDecrement - 1)
+
+    const c = all(a, b)
+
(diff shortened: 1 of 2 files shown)


src/domain-function.test.ts
```diff
@@ -50,7 +50,7 @@ describe('makeDomainFunction', () => {
       envParser,
     )(async ({ id }, { uid }) => [id, uid])
 
-    expect(await handler({ id: '1' }, { uid: '2' })).toEqual({
+    expect(await handler({ id: '1' }, { uid: 2 })).toEqual({
```
gustavoguichard, line 53: couldn't figure out why this line was changed.. mistake?

diogob: The preprocess is not part of the type inference by `z.infer`. Which means that it cannot be relied upon to type-check the code. In practice the consumers of the library would move any environment preprocess to the call site. There seems to be very few cases such as this in our current code bases.

diogob: Maybe I should remove the preprocess from this env to avoid having bogus examples in the tests.

diogob:
I'm closing this for now since it seems to entail a lot of work to be done properly.

PR #19, opened 2022-04-23
Allow a domain function to throw custom error data

danielweinmann:
## Purpose

This PR allows domain functions to throw partial `ErrorData` to be able to return `inputErrors` and `environmentErrors` as well as `errors`. This is useful for implementing validations that go beyond the parsing of the schema, like "Email is already taken", for example.

With this, whenever a server-side validation fails, we can simply throw the appropriate input errors, environment errors, or global errors.

What do you guys think about this solution?


gustavoguichard (commented):
You had me on the tests already. I can see a reason for this feature.

Although I'm not sure if constructing Errors like that might be anti-pattern.. I was just wondering what you @danielweinmann  and @diogob think of having a [custom Error](https://javascript.info/custom-errors) for those cases?

I was just playing in the console tab (without TypeScript) and got this:
Captura de Tela 2022-04-23 às 21 18 49

What do you think of it?

gustavoguichard:
We might even call it `DomainFunctionError` which will also be an instance of `Error`.

src/errors.ts
```diff
@@ -1,5 +1,31 @@
 import * as z from 'zod'
-import type { ErrorWithMessage, SchemaError } from './types'
+import type { ErrorData, ErrorWithMessage, SchemaError } from './types'
+
+function parseError(error: unknown): ErrorData {
```
diogob, line 4: I would rather have narrower types for potential errors since programmers are mostly in control. this would eliminate the need for a `parseError` and simplify the PR  without compromising the DX.

src/errors.ts
```diff
@@ -1,5 +1,31 @@
...
+  }
+
+  return {
+    errors: [toErrorWithMessage(error)],
+    inputErrors: [],
+    environmentErrors: [],
+  }
+}
+
+function isErrorWithErrorData(error: unknown): error is Partial<ErrorData> {
```
diogob, line 20: Instead of typing this as a Partial I would rather have concrete subtypes. This would make the types easier to read and reason about. For instance, today `ErrorData` is defined as:
```ts
type ErrorData = Omit<ErrorResult, 'success'>
```

But we could build up from simpler error types until we get to error result:

```ts
type ErrorData = GeneralErrors & InputErrors & EnvironmentErrors
type ErrorResult = { success: false } & ErrorData
```

diogob: In this case you cloud have custom exceptions like the one suggested by @gustavoguichard but more than one to avoid having to check the presence of fields in runtime. It's best to ensure the shape of the error during type check and reduce our runtime logic.

diogob (changes requested):

diogob (approved):

PR #21, opened 2022-05-02
Update dependencies

gustavoguichard:


diogob:
this is probably stale, we can reopen later.

PR #23, opened 2022-05-31
Add InputErrors exception so we can return multiple input errors at once

### LINKED ISSUE #22: Throw multiple input errors at once
There are use cases such as [this one](https://github.com/SeasonedSoftware/remix-forms/issues/21), where throwing a single exception with multiple input errors is desirable. We should add a custom error constructor for that. Perhaps just have a plural `InputErrors` that will take a non-empty list as argument. If we are fine breaking backwards compatibility we could also just change the `InputError` type.

diogob:
# Purpose

This should close #22

# Technical details

We have decided to add a plural version `InputErrors` so the main use case maintains a simpler API and we also avoid breaking backwards compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,25 @@ failedResult = {
 */
 ```
 
-The same can be done for environment errors by throwing an `EnvironmentError`.
+To throw several input errors in one shot you can use the pluralized version `InputErrors` as in:
+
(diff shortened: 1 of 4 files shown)


danielweinmann (approved):
LGTM

Issue #24, opened 2022-06-08
errorMessagesForSchema function cannot get error messages for nested schema fields


correiarmjoao:
**What version**
0.3.0

**What happend**
errorMessagesForSchema function cannot get error messages for nested schema fields since errorMessageFor only checks the schema keys against the joined path.

https://github.com/SeasonedSoftware/remix-domains/blob/e2288f0d3885bdcd0d10df575f4974f20186014a/src/errors.ts#L22-L38

**Steps to reproduce**
``` 
const data = {
  a: "bar",
  b: "foo",
  c: {
    c1: "c1 foo",
    c2: "c2 bar"
  },
  d: {
    d1: "d1 foo",
    d2: "d2 bar"
  }
}

const schema = z.object({
  a: z.string(),
  b: z.string(),
  c: z.object({
    c1: z.string(),
    c2: z.array(z.string()),
  }),
  d: z.object({
    d1: z.object({
      d1a: z.string(),
      d1b: z.number()
    }),
    d2: z.array(z.string()),
  })
});

const domainFn = makeDomainFunction(schema)(async (data) => {
  return data
})

export const loader = async () => {
  const result = await domainFn(data)
  return errorMessagesForSchema(result.inputErrors, schema)
}
```

**domainFn return**
```
{
  "success": false,
  "errors": [],
  "inputErrors": [
    {
      "path": [ "c", "c2" ],
      "message": "Expected array, received string"
    },
    {
      "path": [ "d", "d1" ],
      "message": "Expected object, received string"
    },
    {
      "path": [ "d", "d2" ],
      "message": "Expected array, received string"
    }
  ],
  "environmentErrors": []
}
```
**errorMessagesForSchema return**
```
{ a: [], b: [], c: [], d: [] }
```

diogob:
@correiarmjoao thanks for the details steps to reproduce. I'm looking into this

gustavoguichard:
Hey, @correiarmjoao ! I believe the [v0.3.1](https://github.com/SeasonedSoftware/remix-domains/releases/tag/v0.3.1) should fix this issue.

Let us know if the problem persists

PR #26, opened 2022-06-15
Implement nested errors in errorMessagesForSchema

diogob:
# Purpose

When a domain function received nested input data, `errorMessagesForSchema` should mirror that structure so we do not lose nested error information.
This should address #24

### DIFF
--- a/src/domain-functions.ts
+++ b/src/domain-functions.ts
@@ -87,6 +87,9 @@ const makeDomainFunction: MakeDomainFunction =
           environmentErrors: [],
         }
       }
+      if (result.success == false) {
+        console.log(JSON.stringify(result.error.format()))
+      }
       return {
         success: false,
         errors: [],
--- a/src/errors.test.ts
+++ b/src/errors.test.ts
@@ -93,4 +93,106 @@ describe('errorMessagesForSchema', () => {
       },
     })
   })
+
+  it('handles nested data errors inside arrays of strings', async () => {
+    const data = {
+      a: 'bar',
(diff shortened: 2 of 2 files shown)


diogob:
@danielweinmann I don't think this will cause any issues for remix-forms, but I haven't done any integration testing. Thoughts?

gustavoguichard (approved):

src/errors.ts
```diff
@@ -24,15 +24,34 @@ const errorMessagesFor = (errors: SchemaError[], name: string) =>
...
 
+type NestedErrors<SchemaType> = {
+  [Property in keyof SchemaType]: string[] | NestedErrors<SchemaType>
+}
+
 const errorMessagesForSchema = <T extends z.AnyZodObject>(
   errors: SchemaError[],
   schema: T,
+  path: string[] = [],
 ) => {
```
gustavoguichard, line 35: Maybe it'd be good to explicitly type the return here at this point

gustavoguichard: For the sake of clarity

gustavoguichard:
@diogob if you feel like waiting for my opinion here I'd ask you to wait for my return on the 29th bc I need to wrap my head around it to understand the implications.

That said I trust you guys to take that decision regardless bc I know some changes must happen as I could see that request coming when we designed the shallow version ;)

diogob:
#31 seems to be a simpler approach that uses only the data contained in `SchemaError` thus avoiding inspection of the schema to generate the nested structure.

PR #27, opened 2022-06-17
fix typo where "object" was written as "objecto"

kory-smith:


diogob:
@kory-smith thanks for the PR

PR #28, opened 2022-06-17
Fix typo

stivaugoin:
I think there is a typo in the example


gustavoguichard:
@stivaugoin thank you for the PR ❤️

PR #31, opened 2022-06-22
Nested errors for schema - take 2

diogob:
# Purpose

When a domain function received nested input data, `errorMessagesForSchema` should mirror that structure so we do not lose nested error information.
This should address #24

## Technical details

This approach seems simpler than the one on #26 since it does not require any info from the schema (other than its type).
Using this code we could keep backwards compatibility or break it just to drop the schema parameter (since we can use just the generic type parameter).

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -33,52 +33,51 @@ const errorMessagesForSchema = <T extends z.AnyZodObject>(
   schema: T,
 ): NestedErrors<z.infer<typeof schema>> => {
   type SchemaType = z.infer<typeof schema>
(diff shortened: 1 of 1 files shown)


jansedlon:
Hi, is there possibility to support zod's discriminated unions? Discriminated unions does not extend the `AnyZodObject` so it's not possible to use it.

diogob:
Hi @jansedlon could you open a separate issue with an example schema?

diogob:
We could probably simplify the code quite a bit, but I want to validate the approach before investing more time.

gustavoguichard (approved):
LGTM

PR #35, opened 2022-09-14
Remove most references to Remix and create a remix folder inside exam…

gustavoguichard:
…ples folder

## Reasoning

This project is only tied to Remix as far as its name. There's no single API in here that needs Remix to run and many other JS projects can benefit from our type-safety, testability and composability, to name few of remix-domains' benefits.

I'm suggesting we name it to domain-functions and foster its usage in other JS ecosystems such as NodeJS, NextJS, Astro, SolidJS, VueJS, Deno / Fresh, CRA - and pretty much any JS framework we can think of.

The plan is to keep populating the examples folder.
If you are watching this repo and feel like trying out remix-domains in other frameworks, feel free to send us a PR adding more examples other than Remix.

Let me hear your thoughts

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,6 @@ It does this by enforcing the parameters' types in runtime (through [zod](https:
 
 ## Benefits
 - End-to-End typesafety all the way from the Backend to the UI
-- Keep your controllers slim and tidy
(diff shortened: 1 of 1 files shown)


diogob (approved):
LGTM

PR #37, opened 2022-09-16
Deno version

diogob:
## Purpose

The idea is to have one version written using Deno and generate the npm package from it.

## TODO

- [x] Migrate all CI scripts
- [x] Ensure npm package is properly generated
- [x] Add documentation about using it from Deno


gustavoguichard:
🤘😍

gustavoguichard:
Before merging we still need to test the build outputs to see if there is going to be any breaking changes

gustavoguichard (changes requested):
Before merging we still need to test the build outputs to see if there is going to be any breaking changes

PR #43, opened 2022-10-18
First combinator

gustavoguichard:
The `first` combinator will return the result of the first successful domain function. It handles inputs and environments like the all function. It is important to notice that all domain functions will be executed in parallel so keep an eye on the side effects.

```ts
const a = makeDomainFunction(
  z.object({ n: z.number(), operation: z.literal('increment') }),
)(async ({ n }) => n + 1)
const b = makeDomainFunction(
  z.object({ n: z.number(), operation: z.literal('decrement') }),
)(async ({ n }) => n - 1)

const result = await first(a, b)({ n: 1, operation: 'increment' })
```


diogob (approved):
Neat!

PR #46, opened 2022-10-28
Replace npm with deno task

diogob:
We were using the root `package.json` just to configure the tasks to run with npm.
This functionality is fully covered by deno tasks using a similar `deno.json` configuration.

I see no reason to keep npm as the project task runner (except when publishing the npm package)


package.json
```diff
@@ -1,27 +0,0 @@
-{
-  "name": "domain-functions",
-  "version": "1.3.0",
-  "description": "Decouple your business logic from your controllers. With first-class type inference from end to end.",
-  "main": "./dist/index.js",
-  "types": "./dist/index.d.ts",
```
gustavoguichard, line 6: Aren't all those lines relevant to the npm ecosystem though?
To find and see a description, to point out where is the entry Js file and the types?

gustavoguichard (changes requested):

package.json
diogob: The `package.json` used for publishing is generated by our [dnt build script](https://github.com/SeasonedSoftware/domain-functions/blob/main/scripts/build-npm.ts#L7):

```bash
❯ deno task build-npm
❯ cat npm/package.json
```

```json
{
  "module": "./esm/index.js",
  "main": "./script/index.js",
  "types": "./types/index.d.ts",
  "name": "domain-functions",
  "version": "1.3.0",
  "description": "Decouple your business logic from your controllers. With first-class type inference from end to end.",
  "license": "MIT",
  "author": "Seasoned Software",
  "bugs": {
    "url": "https://github.com/SeasonedSoftware/domain-functions/issues"
  },
  "homepage": "https://github.com/SeasonedSoftware/domain-functions",
  "exports": {
    ".": {
      "import": "./esm/index.js",
      "require": "./script/index.js",
      "types": "./types/index.d.ts"
    }
  },
  "scripts": {
    "test": "node test_runner.js"
  },
  "dependencies": {
    "qs": "^6.10.3",
    "zod": "^3.19.1"
  },
  "devDependencies": {
    "@types/node": "16.11.37",
    "chalk": "4.1.2",
    "@deno/shim-deno": "~0.6.0",
    "undici": "^5.3.0"
  }
}
```

The only parts of the root `package.json` file being used are the npm scripts and the version number (which I ported to `deno.json`)

gustavoguichard (approved):

PR #47, opened 2022-11-10
Preserve original exception when domain function is handling an unknown thrown value

diogob:
## Purpose

The main use case is to handle exceptions thrown in 3rd party code without having to resort to try/catch.

Let's say you have a library called `ApiClient`. This library is used directly in domain functions, but sometimes throw errors
containing details information on why a certain api call failed.

```ts
const thisMightFail = makeDomainFunction(z.object({ id: z.number() }))(
  async () => {
    const externalData = ApiClient.call('doSomething')
    ...
  },
)
```

Since we don't know what the 3rd party code does, we could capture a failure in the `errors` field of our `ErrorResult`.
However, in the current version, the exception would be discarded and you would only have access to the `message` field of the exception.

The current work-around is to use a `try/catch` within the domain function:

```ts
const thisMightFail = makeDomainFunction(z.object({ id: z.number() }))(
  async () => {
    try {
      const externalData = ApiClient.call('doSomething')
      ...
    } catch (error) {
      throw new Error("Additional error details: " + String(error.data.someNestedDetail))
    }
    ...
  },
)
```

Since we already have a combinator to map errors (`mapError`), it would be handy to have access to all exception data and just map the error:

```ts
const thisMightFail = mapError(makeDomainFunction(z.object({ id: z.number() }))(
  async () => {
    const externalData = ApiClient.call('doSomething')
    ...
  }
  (result: ErrorData) => ({
      ...result,
      errors: result.errors.map((e) => { message: 'Additional error details: ' + String(e.?.exception?.data?.someNestedDetail) })
    })
)
```

The approach enabled by this PR has the advantage of always bubbling up the exception details, so one can always call `mapError` regardless of how the domain function is composed.

## Caveat

I used a reference to the exception instead of cloning the object. Since anything can bee thrown, I decided to keep the type unknown. So we keep the original exception opaque and avoid the complexities and risks of cloning something that we know nothing about.

I consider this a minor inconvenience since mutating the exception in place would be a terrible practice.

## Bonus points

* This enables the user to make error mappers that take advantage of the name in the `Error` type and create a custom domain error from 3rd party APIs. This could extremely useful to handle database errors automatically and turn them into domain errors for instance. 
* This would also enable automatic instrumentation to track full exceptions from domain functions applying `mapError` to already created DFs.

### DIFF
--- a/src/domain-function.test.ts
+++ b/src/domain-function.test.ts
@@ -2,6 +2,7 @@
 import { describe, it } from 'https://deno.land/std@0.156.0/testing/bdd.ts'
 import {
   assertEquals,
+  assertObjectMatch,
   assertRejects,
 } from 'https://deno.land/std@0.160.0/testing/asserts.ts'
 import { z } from 'https://deno.land/x/zod@v3.19.1/mod.ts'
@@ -10,8 +11,13 @@ import {
   mapError,
   makeDomainFunction,
   fromSuccess,
+  map,
+  pipe,
+  all,
+  first,
+  merge,
+  sequence,
 } from './domain-functions.ts'
-import { map, pipe, all, first, merge, sequence } from './domain-functions.ts'
 import {
   EnvironmentError,
   ResultError,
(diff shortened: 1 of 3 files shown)


gustavoguichard:
@diogob  could you give some examples for the bonus points?

diogob:
> @diogob could you give some examples for the bonus points?

Sure, for the first point, imagine that we want to map database lookup failures into a 302 redirect. You could use this information on the loader:

```ts
const result = someDomainFn()

if(result.success) {
  return result.data
} else {
  if(result.errors.some((e) => e?.exception?.name === 'NotFoundError') return redirect ....
}
```

Or you could also use `mapError` to create a set of domain errors.

as for the second point, imagine we have a hypothetical error instrumentation service called `Centry`:

```ts
const errorReporter = (error: ErrorData) => {
  error.errors.forEach((e) => {
    if (e?.exception instanceof Error) {
      CentryApi.sendError(e)
    }
  })
}
const result = mapError(someDomainFn, errorReporter)() 
```

gustavoguichard:
Got it! Beautiful!

gustavoguichard (approved):

PR #49, opened 2022-12-04
Fix bug in "errorMessagesFor" example

yurilaguardia:
This example of the `errorMessagesFor` function returns a string array, but the example in the README was trying to access an unexisting `message` property on the result instead of indexing that array.


gustavoguichard (dismissed):
Great catch! Thank you for the collaboration ❤️

gustavoguichard (approved):

PR #52, opened 2022-12-09
Allow the user to define DFs that will take no input parameters

diogob:
## Purpose

It seems very akward to be forced to pass parameters when they are not going to be used.
Currently a DF without any parameters look something like this:

```ts
const parser = z.undefined()
const handler = makeDomainFunction(parser)(async () => 'no input!')
const result = await handler(undefined)
//                               ^ looks silly
```

this PR makes the input parameter optional:

```ts
const parser = z.undefined()
const handler = makeDomainFunction(parser)(async () => 'no input!')
const result = await handler()
//                          ^ not quite so silly anymore
```

I checked the type inference of the results and compositions and nothing changes after this change is applied.


gustavoguichard:
Do we still have to define a schema tho?
Let's also try to remove that requirement?

diogob:
I just used the `z.undefined()` parser. It makes sense since once you do not give any parser, the function should not take any input.

gustavoguichard (approved):
It's weird why didn't do it before.

jansedlon:
Hey, if you don't pass anything, why do you need domain function at all?

jansedlon:
Fair enough :)

PR #57, opened 2022-12-13
Create a trace function that will capture the inputs and outputs of any DF and call an arbitraty function.

diogob:
## Purpose

This is a more generic version of my earlier proposal of tracing function. 

When compared to `mapError`, the function `trace` cannot affect any output and has access to the input parameters before parsing.

I do believe that this could be in another library, but having this here would already allow us to conveniently create a debugging function as

```ts
const ladyBug = trace(console.trace)
```

It would be trivial also do create a reporting function for external services to keep track of errors:

```ts
const trackErrors = trace(({ input, output, result }) => {
  if(!result.success && someOtherConditions(result)) {
    sendToExternalService({ input, output, result })
  }
})
```


gustavoguichard:
I really like the changes since yesterday ;)

gustavoguichard (approved):
LGTM

README.md
```diff
@@ -296,6 +297,35 @@ errorMessagesForSchema(result.inputErrors, schema)
 */
 ```
 
+### Tracing
+Whenever you need to intercept inputs and the result of a domain function without changing them, there is a function to help you that called `trace`.
```
gustavoguichard, line 301:
```suggestion
Whenever you need to intercept inputs and the result of a domain function without changing them, there is a function called `trace` that can help you..
```

README.md
```diff
@@ -296,6 +297,35 @@ errorMessagesForSchema(result.inputErrors, schema)
...
+
+The most common use case is to log things to the console. Let's say you want to log failed domain functions, you could create a function such as:
```
gustavoguichard, line 303:
```suggestion
The most common use case is to log failures to the console or to an external service. Let's say you want to log failed domain functions, you could create a function such as:
```

gustavoguichard (approved):

PR #58, opened 2022-12-17
Type tests

gustavoguichard:
## Purpose
Since this is a TS heavy library I think it is time for us to introduce type-level tests and avoid breaking type inference on our releases.

## Tech details
- I added 2 new type utilities to help doing type-level testing: `Expect` and `Equal`
- I started by testing the `types.ts` file using namespaces (to repeat types and consts) as if they were `describe` blocks.
- The files that only have type tests must have a dummy runtime test at the bottom.
- I'm using `// @ts-expect-error` whenever I feel it makes sense to expect something to break
- on the `domain-functions.test` file I'm testing **every** composition inference as we've seen code refactorings that would only mess with one or two inferences

## I caught a type bug 🐞
The newly introduced `trace` function kills the type inference of the composition applied to it.
Any DF given to `trace` will be inferred as `DomainFunction<unknown>`.

That was introduced in favor of typing the `result` of the trace callback.
I'm gonna do a PR on top of this one to return the inference to the composition. It'll kill the inference of the `result` but I think the value of having the composition strongly typed is far superior.

### DIFF
--- a/src/domain-function.test.ts
+++ b/src/domain-function.test.ts
@@ -17,7 +17,7 @@ import {
   first,
   merge,
   sequence,
-trace,
+  trace,
 } from './domain-functions.ts'
 import {
   EnvironmentError,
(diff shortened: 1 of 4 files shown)


diogob:
Great PR :tada: 
I just merged the fix in the trace function and tweaked the result type to `Result<unknown>` instead of just `unknown`.

diogob (approved):
Looks great!

PR #61, opened 2022-12-19
Add legacy types back with deprecation messages 👴🏼

gustavoguichard:
To avoid having to bump up the major version of the lib we will keep the deprecated types for now.

Let us know if you were using any of these types.


diogob (approved):
LGTM

PR #62, opened 2022-12-21
Add a little helper to type the DF environments after composition

gustavoguichard:
Just a draft I'd like to discuss ;)


diogob:
Let's move this conversation to #69

PR #63, opened 2023-02-02
docs: punctuation

jly36963:
Not the most substantial PR, mostly a way for me to:
- make an easy first contribution
- get in a copy-editor mindset, in order to read the docs more thoroughly.

Hopefully more meaningful contributions in the future 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Keep your business logic clean with Domain Functions
 
-Domain Functions helps you decouple your business logic from your controllers. With first-class type inference from end to end.
-It does this by enforcing the parameters' types in runtime (through [zod](https://github.com/colinhacks/zod#what-is-zod) schemas) and always wrapping results (even exceptions) into a `Promise<Result<Output>>` type.
+Domain Functions helps you decouple your business logic from your controllers, with first-class type inference from end to end.
+It does this by enforcing the parameters' types at runtime (through [zod](https://github.com/colinhacks/zod#what-is-zod) schemas) and always wrapping results (even exceptions) into a `Promise<Result<Output>>` type.
 
 ![](example.gif)
 
@@ -45,11 +45,12 @@ It does this by enforcing the parameters' types in runtime (through [zod](https:
 - [Acknowlegements](#acknowlegements)
 
 ## Benefits
-- End-to-End typesafety all the way from the Backend to the UI
-- Removes the plumbing of extracting and parsing structured data from your Requests
-- Keep your domain functions decoupled from the framework, with the assurance that your values conform to your types
-- Easier to test and maintain business logic
-- Business Logic can be expressed in the type system
+
+- Provides end-to-end type safety, all the way from the Backend to the UI
+- Removes the "plumbing": Extracting and parsing structured data from your Requests
+- Keeps your domain functions decoupled from the framework, with the assurance that your values conform to your types
+- Facilitates easier testing and maintainence of business logic
+- Allows business logic to be expressed in the type system
 
 ## Quickstart
 
@@ -61,7 +62,7 @@ npm i domain-functions zod
 import { makeDomainFunction, inputFromForm } from 'domain-functions'
 import * as z from 'zod'
 
-const schema = z.object({ number: z.preprocess(Number, z.number()) })
+const schema = z.object({ number: z.coerce.number() })
 const increment = makeDomainFunction(schema)(async ({ number }) => number + 1)
 
 const result = await increment({ number: 1 })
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,7 +62,7 @@ npm i domain-functions zod
 import { makeDomainFunction, inputFromForm } from 'domain-functions'
 import * as z from 'zod'
 
-const schema = z.object({ number: z.preprocess(Number, z.number()) })
+const schema = z.object({ number: z.coerce.number() })
```
jly36963, line 65: Zod now supports [primitive coercion](https://zod.dev/?id=preprocess) without the need for .preprocess()

README.md
```diff
@@ -343,7 +354,8 @@ const c = makeDomainFunction(z.object({ id: z.number() }))(async ({ id }) => Boo
 const results = await all(a, b, c)({ id: 1 })
 ```
 
-On the exemple above, the result will be of type `Result<[string, number, boolean]>`:
+For the exemple above, `data` will be of type `Result<[string, number, boolean]>`:
```
gustavoguichard, line 357: This correction changed the meaning of the original sentence.
Data will be [string, number, boolean] but we want to show that the result of the application of the domain function is an object of type: Result<[string, number, boolean]> . Which might be an ErrorResult without data.

README.md
```diff
@@ -354,7 +366,7 @@ On the exemple above, the result will be of type `Result<[string, number, boolea
 }
 ```
 
-If one or more of the functions fails, the errors will be concatenated:
+If any of the functions fail, the `errors` field will be populated:
```
gustavoguichard, line 369: How to express that the errors of multiple DFs are going to be joined (concatenated)?

README.md
```diff
@@ -419,7 +434,7 @@ if ('resultA' in result.data) return console.log('function A succeeded')
 return console.log('function B succeeded')
 ```
 
-If every domain function fails, the errors will be concatenated:
+If every domain function fails, the `errors` field will be populated:
```
gustavoguichard, line 437: Here we also want to express the errors will be joined (concatenated)

README.md
```diff
@@ -586,7 +610,8 @@ const fetchFullName = pipe(
 const result = fetchFullName({ userId: 2 })
 ```
 
-On the exemple above, the result will be of type `Result<string>` and tis value something like:
+On the exemple above, the result's `data` field will be of type `Result<string>`, and its value something like:
```
gustavoguichard, line 613: "For the example above" for consistency?

gustavoguichard (changes requested):
Great contribution @jly36963 !
Welcome to the repo 🤗 

I left few comments, and blocked the PR due to one correction that changed the meaning of the paragraph.

Overall I'm very happy that someone took the time to read the whole documentation so carefully!

README.md
jly36963: My bad -- I had originally misunderstood the Result type, made similar changes in a few places, realized the mistake, reverted it everywhere else, but missed this one.

README.md
```diff
@@ -739,7 +776,8 @@ async (request: Request) => {
 
 ### inputFromFormData
 
-Extracts a structured object from a `FormData`:
+`inputFromFormData` extracts values from a `FormData` object into an object literal:
```
jly36963, line 779: Let me know if the `object literal` assumption is incorrect, as all the input resolvers have the same return type as [importFromSearch](https://github.com/seasonedcc/domain-functions/blob/ba588c1d2ba054b7dacbcb41f6aae0a86a7873d1/src/input-resolvers.ts#L3)

jly36963: Not sure if `plain object` or some other name might be better

README.md
```diff
@@ -328,13 +337,15 @@ const trackErrors = trace(({ input, output, result }) => {
 ```
 
 ## Combining domain functions
-These combinators are useful for composing domain functions. The returning type of all of them is another `DomainFunction` thus allowing further application in more compositions.
+
+These combinators are useful for composing domain functions. They all return another `DomainFunction`, thus allowing further application in more compositions.
```
jly36963, line 341: Out of curiosity, is there definitive nomenclature for the input DFs as well as the resulting composed DF? There are parts of the docs that would be more clear (and easier to explain) if I could refer to them as `parent`/`children` or similar

gustavoguichard: We never thought about it because we use to see them all as just DF.
I get your point though, maybe we could call it: composite df

README.md
gustavoguichard: I'm not sure about the correctness but I like to think of it as structured objects. There's the Js method structuredClone that implies that it has deep properties. I like to highlight the depth as it is a common request from Remix newcomers learning to deal with FormData, so somehow it'd be cool to let it clear.

gustavoguichard (approved):
LGTM.

@diogob any further comments?

gustavoguichard:
@jly36963 could you squash those commits prior to the merging?

diogob (approved):
Much better

PR #65, opened 2023-02-20
Add a prettify type to make the presented MergeObjects type more read…

gustavoguichard:
…able

I learned this trick few days ago in [this thread](https://twitter.com/GabrielVergnaud/status/1622899119856525313).

It will make the output type of a `merge` combination more readable:

## Before
Screenshot 2023-02-20 at 14 21 34

## After
Screenshot 2023-02-20 at 14 22 07

PR #66, opened 2023-02-20
Add a new `collect` combinator and warn about pitfalls with `merge`

gustavoguichard:
This is not my idea, it was @jly36963 's idea and I loved it.

Still not sure about the name but I think if we had this combinator since the beginning we wouldn't have `merge` or even `all` (... maybe) as this combinator does not enforce an object as a return type of every domain function.

Ideas?

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,8 +22,9 @@ It does this by enforcing the parameters' types at runtime (through [zod](https:
   - [Tracing](#tracing)
 - [Combining domain functions](#combining-domain-functions)
   - [all](#all)
-  - [first](#first)
+  - [namedAll](#namedall)
   - [merge](#merge)
+  - [first](#first)
   - [pipe](#pipe)
   - [sequence](#sequence)
   - [map](#map)
@@ -40,7 +41,6 @@ It does this by enforcing the parameters' types at runtime (through [zod](https:
   - [inputFromFormData](#inputfromformdata)
   - [inputFromUrl](#inputfromurl)
   - [inputFromSearch](#inputfromsearch)
-- [Need help?](#need-help)
 - [Resources](#resources)
 - [Acknowlegements](#acknowlegements)
 
@@ -389,6 +389,79 @@ const results = await all(a, b)({ id: 1 })
 }*/
 ```
 
+### namedAll
+
+`namedAll` works like the `all` function but receives its constituent functions inside a record with string keys that identify each one. The shape of this record will be preserved for the `data` property in successful results.
+
+The motivation for this is that an object with named fields is often preferable to long tuples, when composing many domain functions.
+
+```ts
+const a = makeDomainFunction(z.object({}))(async () => '1')
+const b = makeDomainFunction(z.object({}))(async () => 2)
+const c = makeDomainFunction(z.object({}))(async () => true)
+
+const results = await namedAll({ a, b, c })({})
+```
+
+For the example above, the result type will be `Result<{ a: string, b: number, c: boolean }>`:
+
+```ts
+{
+  success: true,
+  data: { a: '1', b: 2, c: true },
+  errors: [],
+  inputErrors: [],
+  environmentErrors: [],
+}
+```
+
+As with the `all` function, in case any function fails their errors will be concatenated.
+
+### merge
+
+`merge` works exactly like the `all` function, except __the shape of the result__ is different.
+Instead of returning a tuple, it will return a merged object which is equivalent to:
+```ts
+map(all(a, b, c), mergeObjects)
+```
+
+The resulting data of every domain function will be merged into one object. __This could potentially lead to values of the leftmost functions being overwritten by the rightmost ones__.
+
+```ts
+const a = makeDomainFunction(z.object({}))(async () => ({
+  resultA: 'string',
+  resultB: 'string',
+  resultC: 'string',
+}))
+const b = makeDomainFunction(z.object({}))(async () => ({ resultB: 2 }))
+const c = makeDomainFunction(z.object({}))(async () => ({ resultC: true }))
+
+const results = await merge(a, b, c)({})
+```
+
+For the example above, the result type will be `Result<{ resultA: string, resultB: number, resultC: boolean }>`:
+
+```ts
+{
+  success: true,
+  data: { resultA: 'string', resultB: 2, resultC: true },
+  errors: [],
+  inputErrors: [],
+  environmentErrors: [],
+}
+```
+
+__Be mindful of__ each constituent domain function's return type. If any domain function returns something other than an object, the composite domain function will return an `ErrorResult`:
+
+```ts
+{
+  success: false,
+  errors: [{ message: 'Invalid data format returned from some domain functions' }],
+  inputErrors: [],
+  environmentErrors: [],
+}
+```
+
 ### first
 
 `first` will create a composite domain function that will return the result of the first successful constituent domain function. It handles inputs and environments like the `all` function.
@@ -455,44 +528,6 @@ const result = await first(a, b)({ id: 1 })
   inputErrors: [],
   environmentErrors: [],
 }*/
-```
-
-### merge
-
-`merge` works exactly like the `all` function, except __the shape of the result__ is different.
-Instead of returning a tuple, it will return a merged object.
-
-The motivation for this is that an object with named fields is often preferable to long tuples, when composing many domain functions.
-
-```ts
-const a = makeDomainFunction(z.object({}))(async () => ({ resultA: '1' }))
-const b = makeDomainFunction(z.object({}))(async () => ({ resultB: 2 }))
-const c = makeDomainFunction(z.object({}))(async () => ({ resultC: true }))
-
-const results = await merge(a, b, c)({})
-```
-
-For the example above, the result type will be `Result<{ resultA: string, resultB: number, resultC: boolean }>`:
-
-```ts
-{
-  success: true,
-  data: { resultA: '1', resultB: 2, resultC: true },
-  errors: [],
-  inputErrors: [],
-  environmentErrors: [],
-}
-```
-
-__Be mindful of__ each constituent domain function's return type. If any domain function returns something other than an object, the composite domain function will return an `ErrorResult`:
-
-```ts
-{
-  success: false,
-  errors: [{ message: 'Invalid data format returned from some domain functions' }],
-  inputErrors: [],
-  environmentErrors: [],
-}
 ```
 
 ### pipe
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ It does this by enforcing the parameters' types at runtime (through [zod](https:
   - [Tracing](#tracing)
 - [Combining domain functions](#combining-domain-functions)
   - [all](#all)
-  - [namedAll](#namedall)
+  - [collect](#collect)
   - [merge](#merge)
   - [first](#first)
   - [pipe](#pipe)
@@ -389,9 +389,9 @@ const results = await all(a, b)({ id: 1 })
 }*/
 ```
 
-### namedAll
+### collect
 
-`namedAll` works like the `all` function but receives its constituent functions inside a record with string keys that identify each one. The shape of this record will be preserved for the `data` property in successful results.
+`collect` works like the `all` function but receives its constituent functions inside a record with string keys that identify each one. The shape of this record will be preserved for the `data` property in successful results.
 
 The motivation for this is that an object with named fields is often preferable to long tuples, when composing many domain functions.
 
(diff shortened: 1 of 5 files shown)


src/types.ts
```diff
@@ -45,6 +45,10 @@ type UnpackAll<List, output extends unknown[] = []> = List extends [
   ? UnpackAll<rest, [...output, first]>
   : output
 
+type UnpackDFObject<Obj extends Record<string, DomainFunction>> =
+  | { [K in keyof Obj]: UnpackData<Obj[K]> }
+  | never
```
gustavoguichard, line 50: As in the #65 this `| never` will prettify the resulting type of the composition.

jly36963:
Obrigado amigo 👍

diogob:
very neat idea!

Since it's very similar to merge while being a total function, how about calling this combinator `safeMerge`?

Actually I feel like deprecating `merge` in favour of this one (since partial functions are a pain).

If we decide to deprecate `merge` the best course might be to use both functions under an overloaded signature and document only this one adding deprecation notes to the original syntax.

gustavoguichard:
@diogob I love the idea! 
I'll tweak the PR a little bit and feel free to add to it!

jly36963:
IMO the functionality is loosely related to `mapValues`

### Examples

lodash `mapValues`

```js
import { mapValues } from 'lodash-es'

const callableMap = {
  a: () => 1,
  b: () => 2
}

mapValues(callableMap, f => f())

// { a: 1, b: 2 }
```

Rust HashMap map
```rs
use std::collections::HashMap;

fn main() {
    let hm: HashMap<&str, i32> = HashMap::from([("a", 1), ("b", 2), ("c", 3)])
        .iter()
        .map(|(k, v)| (*k, *v * 2))
        .collect();
    for (k, v) in &hm {
        println!("{}: {}", k, v);
    }
}
```

I don't know Haskell, but Data Map has something like `mapKeys` or `mapValues`, right?

###  mapValues as a name

I know the `map` name is already used for mapping result data, and it's similar to Rust's Option or Result `map` (map the value if Some/Ok, do nothing if None/Error). 

Would a name like `mapValues` be too generic (as its specifically mapping callables to their results), or too confusing (as `map` already serves a different purpose?

gustavoguichard:
@jly36963 I think it would be confusing.
IMO Lodash should be one of the last place to get inspiration for naming conventions as the `map` word has an entire different meaning there.

I like the idea of deprecating the current `merge` in favor of this one. I'll post a PR for that rn so we can discuss the possibility.

diogob:
I have renames the PR for easier future reference.

src/index.ts
```diff
@@ -1,4 +1,15 @@
-export * from './domain-functions.ts'
+export {
```
diogob, line 1: If we drop the `oldMerge` export this change would become optional. Having said that, having an explicit list of exports does not bother me one bit.

danielweinmann:
With this new merge, how would I achieve the following? 

```ts
const firstDF = makeDomainFunction()(async () => ({ items: [1, 2, 3] }))
const secondDF = makeDomainFunction()(async () => ({ currentUser: { id: 1 } }))
```

Then, I want the data to be:

```ts
{
  items: [1, 2, 3],
  { currentUser: { id: 1 } }
}
```

Would I need to `merge` and then `map`? If so, I don't like deprecating the old merge. This use case is pretty common in my day-to-day.

src/index.ts
gustavoguichard: I thought the same. I think we can reduce the diff by not exporting the old merge.

gustavoguichard:
@danielweinmann you'd probably need to use a combination of all, map and mergeObjects:

```ts
const { resultA, resultB } =
  map(all(dfA, dfB), mergeObjects)
```

danielweinmann:
> @danielweinmann you'd probably need to use a combination of `all`, `map` and `mergeObjects`:
> 
> ```ts
> const { resultA, resultB } =
>   map(all(dfA, dfB), mergeObjects)
> ```

Then why get rid of the current merge? Its introduction was a game-changer for me. I can't see myself going through the loops and bounds of all + map + mergeObjects. I'd probably go back to using regular functions instead of composing DFs.

gustavoguichard:
> Then why get rid of the current merge? Its introduction was a game-changer for me. I can't see myself going through the friction of all + map + mergeObjects. I'd probably go back to using regular functions instead of composing DFs.

I think we wouldn't have created it if we had the implementation proposed in this PR. So we wouldn't need to enforce objects as return types.

There is no problem of keeping both, though.. the issue IMO is the naming.

FWIW this would be a valid drop-in replacement for the current `merge` that pass every test and type-checks:
```ts
const oldMerge = <Fns extends DomainFunction[]>(
  ...fns: Fns
): DomainFunction<MergeObjs<UnpackAll<Fns>>> => map(all(...fns), mergeObjects)

const c = oldMerge(firstDF, secondDF)
//    ^? DomainFunction<{ items: number[]; currentUser: { id: number } }>
```

danielweinmann:
What use cases do you have in mind for the new merge where we name the keys at the composition-level, @gustavoguichard? Most use cases I can think of name the keys at each DF.

gustavoguichard:
I see 2 main benefits:
- Stop enforcing objects as return types - which is something that always bugged me
- Avoiding conflicts / possible overwrites - which will likely decrease cognitive load on big codebases

But I get you, I don't see myself not using `oldMerge` at all, I was able to avoid conflicts and will probably keep using it.

I want to flip the question though, @danielweinmann :

Do you see benefits and use cases for this new combinator?
If we could freely change the API, how would you name both of them?

gustavoguichard:
There's also the possibility of leaving the function polymorphic without the deprecation messages.

😅 I'm not a fan of the idea but wanted to put it here for discussion.

jly36963:
@danielweinmann FWIW, here are my two cents.

### Namespace

Guga gave me a quick catch-up on the history of DF. He mentioned that situations where `all` returns long tuples effectively created a need for namespacing. Naturally, a pattern of returning objects (for destructuring) and using `merge` to merge the resulting data objects made sense.

I understand why you wouldn't want to break existing behavior, especially when people might rely heavily on the current merge API. the `merge` name implies merging objects, which the proposed behavior _doesn't_ do.

Would a better name for the new merge paradigm be `many` or `named`? One would be executing many named queries.

### Why the new method

For the sake of this example, let's call the new behavior `named`

IMO, it's a simpler mental model for a function to return what it's name describes (without the added layer of nesting):

```ts
const getUser = async (id) => ({ id: 1 });
const getItems = async (id) => [1,2,3];

const data = { user: await getUser(1), items: getItems(1) }
```

So in terms of DF:

```ts
const getItems = makeDomainFunction()(async () => [1, 2, 3]);
const getUser = makeDomainFunction()(async () => ({ id: 1 }));

const result = await named({ items: getItems, user: getUser });

/*

{
  success: true,
  data: { items: [1,2,3], currentUser: { id: 1 } },
  inputErrors: [],
  environmentErrors: [],
  errors: [],
}
*/
```

Do you think this is sensible?

diogob:
> There's also the possibility of leaving the function polymorphic without the deprecation messages.
> 
> sweat_smile I'm not a fan of the idea but wanted to put it here for discussion.

I would rather keep different names if we are not deprecating the merge combinator.
I like @jly36963 's `named` suggestion since it is more descriptive of the intent.

@danielweinmann @gustavoguichard I don't mind keeping both. The reason why I suggested deprecating merge is because there are parameters that will type-check but generate a run-time error. This increases cognitive load and creates a pitfall for newcomers, since we could easily introduce a merge that breaks the code in a chain of compositions.  

However, after reading all these comments I realize that the old merge compelled users to write their return types in a certain shape. And rewriting all domain functions in a big project would be painful and wasteful,  besides, having both does not add significant complexity to the project. But using the overload indefinitely would bee too confusing since we would have a partial function and a total one with slightly different semantics sharing a name, a recipe for subtle bugs in runtime.  

I propose: 

* We keep `merge` as is. 
* Rollback the deprecation.
* Add the new combinator calling it `named`.

danielweinmann:
> > There's also the possibility of leaving the function polymorphic without the deprecation messages.
> > sweat_smile I'm not a fan of the idea but wanted to put it here for discussion.
> 
> I would rather keep different names if we are not deprecating the merge combinator. I like @jly36963 's `named` suggestion since it is more descriptive of the intent.
> 
> @danielweinmann @gustavoguichard I don't mind keeping both. The reason why I suggested deprecating merge is because there are parameters that will type-check but generate a run-time error. This increases cognitive load and creates a pitfall for newcomers, since we could easily introduce a merge that breaks the code in a chain of compositions.
> 
> However, after reading all these comments I realize that the `merge` compelled users to write their return types in a certain shape. And rewriting all domain functions in a big project would be painful and wasteful, besides, having both does not add significant complexity to the project. But using the overload indefinitely would bee too confusing since we would have a partial function and a total one with slightly different semantics sharing a name, a recipe for subtle bugs in run-time.
> 
> I propose:
> 
> * We keep `merge` as is.
> * Rollback the deprecation.
> * Add the new combinator calling it `named`.

I like that! What about calling the new one `namedAll`? After giving it some though, it has a lot more in common (in term of use cases) with `all` than with `merge`. It feels a little more intuitive to make the name analogous to `all`.

gustavoguichard:
> I like that! What about calling the new one namedAll? After giving it some though, it has a lot more in common (in term of use cases) with all than with merge. It feels a little more intuitive to make the name analogous to all.

I like it better than `named` even though I like to stick to one word per method. This one is quite short though.

Do you guys still think we should remove `merge` from the README anyway? The points exposed (partial function and the possibility of conflicts) still apply and I'm not sure about encouraging them.

danielweinmann:
> > I like that! What about calling the new one namedAll? After giving it some though, it has a lot more in common (in term of use cases) with all than with merge. It feels a little more intuitive to make the name analogous to all.
> 
> I like it better than `named` even though I like to stick to one word per method. This one is quite short though.
> 
> Do you guys still think we should remove `merge` from the README anyway? The points exposed (partial function and the possibility of conflicts) still apply and I'm not sure about encouraging them.

I _really_ want to encourage the usage of `merge`. It's the best combinator for when your combining DFs that return namespaced objects (which are 90% of my use cases). 

There are drawbacks to the type-safety of other combinators, e.g. `pipe` not guaranteeing the return type of the first function is a correct input for the second. To me, most combinators are sharp knives that need to be handled with care.

Going back to the naming brainstorm, how do you feel about `record`?

gustavoguichard:
@danielweinmann check my last commit's diff. I think I made it clear in the README about the potential pitfalls.

I like `record`... still not sure about it 😅
@danielweinmann, @jly36963, @diogob  let's make a quick poll: react with 👍 for `namedAll` and 😀 for `record`

gustavoguichard:
FWIW in [f8c6df8](https://github.com/seasonedcc/domain-functions/pull/66/commits/f8c6df89a8eb2502ad43861ee138b43cd16c2377) I fixed `MergeObjs` type so it behaves according to what I documented and to the actual runtime.

I also added tests for that!

jly36963:
Initially I liked `record`, but I have one concern. [TS Record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) implies that the keys are all of one type and that the values are all another. In this case, the return types of the constituent DFs can be anything, so the latter assumption about the values sharing a type isn't really true.

As @danielweinmann said, its behavior is closely related to `all`. I'm totally cool with `namedAll`

djkz:
Love the functions, would like to suggest another name for it: `collect` 

Since you are basically collecting the results of a bunch of domain functions into a single result I think it makes more sense on what it does than "named".

Any thoughts on that name?

(Another suggestion would be something like `build` as in building the object)

danielweinmann:
> Initially I liked `record`, but I have one concern. [TS Record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) implies that the keys are all of one type and that the values are all another. In this case, the return types of the constituent DFs can be anything, so the latter assumption about the values sharing a type isn't really true.
> 
> Calling it `record` might be misleading IMO.
> 
> As @danielweinmann said, its behavior is closely related to `all`. I'm totally cool with `namedAll`

Great point, @jly36963! I agree that record is not precise enough 👌🏼

danielweinmann:
> Love the functions, would like to suggest another name for it: `collect`
> 
> Since you are basically collecting the results of a bunch of domain functions into a single result I think it makes more sense on what it does than "named".
> 
> Any thoughts on that name?
> 
> (Another suggestion would be something like `build` as in building the object)

`build` is interesting! All the years of `FactoryBot`ing made me associate build with creating an object :D

jly36963:
I like `collect` 👍

README.md
```diff
@@ -389,6 +389,79 @@ const results = await all(a, b)({ id: 1 })
 }*/
 ```
 
+### namedAll
```
diogob, line 392: we should update the README with the new `collect` name

diogob (approved):
I just added a commit renaming the combinator in the README.

LGTM

PR #67, opened 2023-02-20
Deprecate oldMerge usage and add a function overload to make the new …

gustavoguichard:
…merge the new merge API

As suggested by @diogob in #66 I added an overload to the `merge` function (previously called `combine` in #66).

When using the old api the function will go through the `oldMerge` method.

Also, with the TS deprecation message the DX is as good as expected, you'll see a line-through on the `merge` function when trying to use the old API but every test and type test will keep succeeding.

tweet2

### DIFF
--- a/README.md
+++ b/README.md
@@ -459,41 +459,31 @@ const result = await first(a, b)({ id: 1 })
 
 ### merge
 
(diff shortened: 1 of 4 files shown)


diogob:
I'm just merging this onto #66 to review everything in one shot

PR #68, opened 2023-02-21
Fix MergeObjs type to behave more like JS's Object.assign and add som…

gustavoguichard:
…e regression tests.

## Reasoning
There was a mismatch between type and runtime behavior of `mergeObjects` as shown here:
```ts
const a = { a: 1, b: 2 }
const b = { b: true }
const c = mergeObjects([a, b]) // should behave like { ...a, ...b }
//    ˆ?  { a: number, b: never }
```

This PR fixes it according to what we would expect: `{ a: number, b: boolean }`. It also enforces the correct parameter types with TS: `DomainFunction<string, unknown>[]`

This bug was mentioned in #66

### DIFF
--- a/src/domain-function.test.ts
+++ b/src/domain-function.test.ts
@@ -507,7 +507,11 @@ describe('merge', () => {
 
   it('should combine many domain functions into one', async () => {
     const a = makeDomainFunction(z.object({ id: z.number() }))(
-      async ({ id }) => ({ resultA: String(id) }),
+      async ({ id }) => ({
+        resultA: String(id),
+        resultB: String(id),
+        resultC: String(id),
+      }),
     )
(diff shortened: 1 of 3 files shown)


src/domain-function.test.ts
```diff
@@ -507,7 +507,11 @@ describe('merge', () => {
 
   it('should combine many domain functions into one', async () => {
     const a = makeDomainFunction(z.object({ id: z.number() }))(
-      async ({ id }) => ({ resultA: String(id) }),
+      async ({ id }) => ({
+        resultA: String(id),
+        resultB: String(id),
+        resultC: String(id),
+      }),
```
gustavoguichard, line 514: This would make the type like: `DomainFunction<never>` without the fix

diogob (approved):
Great catch, this long weekend has been productive :P

PR #69, opened 2023-02-22
Carry parser types

diogob:
This PR works on ideas of #17 and #62 .

## Motivation

* Typing input and environment in DFs is not backwards compatible and reduces DFs usefulness while increasing their safety.
* Therefore if we could keep the current signature for DFs but have TS tell us in compile time if a composition will work seem to be an optimal solution.
* It would also be nice to give some helper that could convert a DF into a DF-like function with a type-safe environment (and perhaps a helper for a fully type-safe function).

## Requirements

* We should be able to carry the types `InputSchema` and `EnvSchema` with the type of each DF to calculate when the compositions work.
* The solution should be backwards compatible.
* All combinators that would be applied on DFs with incompatible inputs or outputs will either fail to type-check or resolve to a `void` output type.

## Current approach

The parameters `Input` and `Environment` where added to the `DomainFunction` generic (see [Phantom types](https://wiki.haskell.org/Phantom_type)). This will allow us to retrieve them when type-checking compositions without interfering in the domain functions signature. It also avoids changes to the run-time data structures since the parsers are known at compile time. With proper defaults for these new parameters we should be able to keep things backwards compatible.

## TODO

* Ensure parallel compositions carry proper composed input parser resulting types:
  * [x] `all`
  * [ ] `collect`
  * [ ] `first`
* Ensure sequential compositions type-check intermediary results and carry proper input parser types:
  * [x] `pipe`
  * [ ] `sequence`
* [ ] Ensure `merge` type checks that all results are mergeable

### DIFF
--- a/src/all.test.ts
+++ b/src/all.test.ts
@@ -6,8 +6,8 @@ import {
 import { z } from 'https://deno.land/x/zod@v3.19.1/mod.ts'
 
 import { makeDomainFunction } from './constructor.ts'
-import { all } from './domain-functions.ts'
-import type { DomainFunction } from './types.ts'
+import { all, strict } from './domain-functions.ts'
+import type { DomainFunction, StrictDomainFunction } from './types.ts'
 import type { Equal, Expect } from './types.test.ts'
 
 describe('all', () => {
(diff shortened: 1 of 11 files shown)


src/types.ts
```diff
@@ -62,6 +66,13 @@ type Prettify<T> = {
 
 type TupleToUnion<T extends unknown[]> = T[number]
 
+type TupleToIntersection<
+  T extends unknown[],
+  output extends unknown = unknown,
+> = T extends [infer first, ...infer rest]
+  ? TupleToIntersection<rest, output & first>
+  : Prettify<output>
+
```
gustavoguichard, line 89: How is this different from `MergeObjs` ?

diogob: The result of `TupleToIntersection<[number, string]>` is `never`. `MergeObjs` currently assumes the input types always have keys.

diogob:
> I like where this is going and just want to leave a note here that we should probably cut a major release after this is merged and remove the deprecated code.

Sure, this certainly will be a breaking change since some old compositions will no longer type-check.

DF 2.0 :sunglasses:

src/all.test.ts
```diff
@@ -133,4 +133,61 @@ describe('all', () => {
...
+
+    type _R = Expect<
+      Equal<
+        typeof d,
+        StrictDomainFunction<[string, number, boolean], { id: number }, string>
+      >
+    >
+
+    //@ts-expect-error: The code below does not type check since our environment parser won't take a number
+    let results = await d({ id: 1 }, 1)
```
gustavoguichard, line 182: isn't the test supposed to fail with envErrors tho??

src/strict.test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    const handler = makeDomainFunction(
+      parser,
+      environmentParser,
+    )(async ({ id }) => id)
+
+    const strictHandler = strictEnvironment(handler)
+
+    strictHandler(
+      'input is unknown, will error on parse',
```
gustavoguichard, line 35: This line shouldn't type check, right? It doesn't comply with `{ id: unknown }`

diogob: just the environment is strict in this version. It uses the `strictEnvironment`

gustavoguichard: Got it!

diogob:
Closing this since it is a bit stale already and we have decided to leave it out of 2.0.
Also, we are still not sure if this feature is desirable to the point of justifying the added complexity.

gustavoguichard:
This work is complete with the release of `composable-functions`! 🎉

PR #70, opened 2023-02-27
Fix npm package for webpack

diogob:
## Fix npm package for webpack

The conditional syntax for `package.json` on [webpack documentation](https://webpack.js.org/guides/package-exports/#conditional-syntax) requires the `default` property to be the last one in the object. The [dnt version](https://deno.land/x/dnt@0.31.0/mod.ts) we were using generated this "exports" property in the package.json:

```json
  "exports": {
    ".": {
      "import": {
        "default": "./esm/index.js",
        "types": "./types/index.d.ts"
      },
      "require": {
        "default": "./script/index.js",
        "types": "./types/index.d.ts"
      }
    }
  },

```

This would break webpack rendering impossible to use domain functions.
A message error would be displayed such as:

```
Module not found: Default condition should be last one
> 1 | import { makeDomainFunction } from "domain-functions";
```

After the update in this PR the issue is fixed with a different generated property syntax:

```json
  "exports": {
    ".": {
      "import": {
        "types": "./types/index.d.ts",
        "default": "./esm/index.js"
      },
      "require": {
        "types": "./types/index.d.ts",
        "default": "./script/index.js"
      }
    }
  },
```


gustavoguichard (approved):
This is so f***d up! I'm glad you took the time to solve it ❤️

PR #71, opened 2023-03-21
Accept sync functions

gustavoguichard:
This always had bugged me. Sometimes we want sync functions as domain-function's body but due to a lack of TS prowess we haven't achieved it yet... well, that time has come =)

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ import { makeDomainFunction, inputFromForm } from 'domain-functions'
 import * as z from 'zod'
 
 const schema = z.object({ number: z.coerce.number() })
-const increment = makeDomainFunction(schema)(async ({ number }) => number + 1)
+const increment = makeDomainFunction(schema)(({ number }) => number + 1)
 
 const result = await increment({ number: 1 })
 /*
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -428,12 +428,12 @@ map(all(a, b, c), mergeObjects)
...
 ```ts
-const a = makeDomainFunction(z.object({}))(async () => ({
+const a = makeDomainFunction(z.object({}))(() => ({
   resultA: 'string',
   resultB: 'string',
   resultC: 'string',
 }))
-const b = makeDomainFunction(z.object({}))(async () => ({ resultB: 2 }))
+const b = makeDomainFunction(z.object({}))(() => ({ resultB: 2 }))
 const c = makeDomainFunction(z.object({}))(async () => ({ resultC: true }))
```
gustavoguichard, line 437: I'll leave one async to show how they can be combined

src/domain-functions.ts
```diff
@@ -94,7 +94,7 @@ function makeDomainFunction<
           ? []
           : formatSchemaErrors(envResult.error.issues),
       }
-    } as DomainFunction<Output>
```
gustavoguichard, line 97: source

PR #72, opened 2023-04-13
Split tests and assert inferred types with more precision

gustavoguichard:
I'm splitting the tests into one file per combinator so it is easier to navigate in the test files.
I also moved the `makeDomainFunction` method to its own file.

Along with that I'm bringing back the `Expect<Equals<T,U>>` type tests because they are more assertive, e.g: this wouldn't pass:
```ts
type _R = Expect<Equal<{ a: string } & { b: number }, { a: string, b: number }>>
```

And by assuring a more deterministic type we are less likely gonna introduce breaking type inference with code changes.


src/domain-function.test.ts
```diff
@@ -1,1244 +0,0 @@
...
-
-  it('returns the error when the mapping function fails', async () => {
-    const a = makeDomainFunction(z.object({ id: z.number() }))(
-      async ({ id }) => id + 1,
-    )
-    const b = () => {
-      throw 'failed to map'
-    }
-
-    const c: DomainFunction<number> = mapError(a, b)
```
gustavoguichard, line 1178: This type was wrong. With the new type assertion tests I was able to catch it:
```ts
Expect<Equal<typeof c, DomainFunction<never>>>
```

diogob (approved):
I wasn't expecting this so soon! Amazing :sunglasses:

Issue #81, opened 2023-07-26
mergeObjects has broken type definition


jam-fran:
It appears that something's broken with `mergeObjects`' type definition in the generated `utils.d.ts` file. I've pasted portion of it below (only included a tiny sample, in the actual file it's thousands and thousands of lines long). In my IDE importing the function crashes the Typescript server and breaks all inferred types. 

[...]

PR #83, opened 2023-07-31
Remove prettify to avoid circular bug

gustavoguichard:
This was reported on #81 , the `Prettify` type is supposed to help with readability of resulting type of `MergeObj` but it has a - now - known bug that can cause infinitely recursive types thus killing TSC.

You can read more about it here: https://github.com/remix-run/remix/pull/6736

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -49,11 +49,10 @@ type UnpackDFObject<Obj extends Record<string, DomainFunction>> =
   | { [K in keyof Obj]: UnpackData<Obj[K]> }
   | never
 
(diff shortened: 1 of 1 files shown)


gustavoguichard:
After investigating a little longer I was able to generate a cleaner version of the `MergeObjs` that won't create the problem.
The file described by @jam-fran is this one: https://www.unpkg.com/domain-functions@1.7.0/types/utils.d.ts

In that file we can see the problem. After making this small change the generated `utils.d.ts` won't carry that bug.

PR #84, opened 2023-07-31
Remove `qs` dependency

gustavoguichard:
I started drafting this PR with the [help of ChatGPT](https://chat.openai.com/share/82b18fe6-5bc3-4f9f-ac13-ec7519d9f993). I didn't analyze the code yet and didn't refactor but the tests are succeeding.

I'll leave this PR as a draft and whenever anyone of us have time we can use it as a base to finally deliver this code 🚀

It should close #77

PR #85, opened 2023-08-02
remove qs take 2

### LINKED ISSUE #77: Remove `qs` as dependency of this library
It doesn't add much value, we could roll out our own helper and `qs` has other dependencies we don't want to be bundling with domain-functions.

diogob:
Another attempt at #84, it should also close #77.

Just trying a different approach for fun.

The previous attempt would still fail in some of the new test cases introduced here.


diogob:
@danielweinmann please take a look at the test cases. Since this is going to be a breaking change let's ensure we cover all cases we care about.

gustavoguichard (approved):

gustavoguichard:
@diogob I changed the base branch for this PR from `main` to `v2`

gustavoguichard:
@diogob , let's get this merged? Can't stand qs anymore:
https://bundlephobia.com/package/domain-functions@1.8.0

danielweinmann (changes requested):
@gustavoguichard, I asked @diogob for a few extra test cases to make sure we're dealing correctly with deeply nested objects/arrays.

diogob:
@danielweinmann @gustavoguichard I have rebased `v2` and this branch on top of the new `v2`.
I have also added a fancy new test case to cover several cases in the same call.

danielweinmann (approved):

PR #87, opened 2023-08-06
Rewrite collect to leverage simpler combinators

diogob:
It seems that `collect` could easily be expressed in terms of `all`.
The parallel application is the same, and we just need to push the keys from the input to the output.

This ensures we never have issues with the output shave as we do in the `merge` combinator.

### DIFF
--- a/src/domain-functions.ts
+++ b/src/domain-functions.ts
@@ -50,7 +50,7 @@ function collect<Fns extends Record<string, DomainFunction>>(
   fns: Fns,
 ): DomainFunction<UnpackDFObject<Fns>> {
   const dfsWithKey = Object.entries(fns).map(([key, df]) =>
-    map(df, (result) => Object.fromEntries([[key, result]])),
+    map(df, (result) => ({ [key]: result })),
   )
   return map(all(...dfsWithKey), mergeObjects) as DomainFunction<UnpackDFObject<Fns>>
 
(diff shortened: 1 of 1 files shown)


src/domain-functions.ts
```diff
@@ -49,40 +49,11 @@ function all<Fns extends DomainFunction[]>(
...
 ): DomainFunction<UnpackDFObject<Fns>> {
-  return async (input, environment) => {
-    const results = await Promise.all(
-      Object.entries(fns).map(
-        async ([key, fn]) =>
-          [key, await (fn as DomainFunction)(input, environment)] as const,
-      ),
-    )
+  const dfsWithKey = Object.entries(fns).map(([key, df]) =>
+    map(df, (result) => Object.fromEntries([[key, result]])),
```
gustavoguichard, line 53: I think this is clear - as in: it is explicit it is an object with only one property 
```suggestion
    map(df, (result) => ({ [key]: result })),
```

gustavoguichard (approved):
Great refactoring

src/domain-functions.ts
```diff
@@ -49,40 +49,11 @@ function all<Fns extends DomainFunction[]>(
...
+  )
+  return map(all(...dfsWithKey), mergeObjects) as DomainFunction<UnpackDFObject<Fns>>
```
gustavoguichard, line 55: Maybe `merge` was a mistake, but we can still...
```suggestion
  return merge(...dfsWithKey) as DomainFunction<UnpackDFObject<Fns>>
```

diogob: true, but the merge definition is more complex (and rewriting that would bring little benefit). I rather leave the map + all since it is a good example of using simpler combinators to build more complex functionality

PR #88, opened 2023-08-06
Add README note steering users away from merge

diogob:
I do not like having partial functions in the same namespace as all the others.
I would leve to move `merge` to another module called `unsafe` or `partial`, or perhaps even deprecate the combinator for future removal.

For now I'm happy just telling visitors to steer clear.


gustavoguichard (approved):
I'm ok about deprecating it too

PR #89, opened 2023-08-06
Refactors some methods such as `merge`, `pipe`, and adds a `safeResult` helper to turn everything into a Result

gustavoguichard:
Inspired by #87 I'm also doing some refactoring on `merge` and `pipe` methods =)


diogob (approved):
Funny, I was just about to write this `pipe` implementation.
I also like the safe result, it make the constructor easier to read, the error cases using the exceptions are also neat.

PR #92, opened 2023-08-07
BREAKING CHANGE: Remove `merge`  method

gustavoguichard:
To prepare for v2 we are removing the `merge` method in favor of `collect` as the latter is a total function while `merge` is a partial one and may cause weird bugs or data loss in compositions.

Keep in mind `merge` can be done out of the existing combinators:
```ts
// Before:
import { merge } from 'domain-functions'

merge(dfA, dfB)

// After:
import { map, all, mergeObjects } from 'domain-functions'

map(all(dfA, dfB), mergeObjects)
```


diogob (approved):
:smiling_face_with_tear:

PR #94, opened 2023-08-10
Removes runtime validation from merge to make its definition shorter

gustavoguichard:
On the path to v2 we decided to keep the `merge` function in the lib with all due warnings.
This will avoid breaking changes.

I removed the runtime validation from the function definition to make it as thin as we think it should be.
It shouldn't add any runtime changes as the validation would be already throwing in case the `merged` was misued. ;)

Important to notice the TS type check will still enforce the correct usage as the parameters must extend `DomainFunction<Record<string, unknown>>`.


diogob (approved):
LGTM

PR #95, opened 2023-08-10
BREAKING CHANGE: Removes `errorMessagesForSchema` utility

gustavoguichard:
This is a breaking change that will go on v2.

We are removing support for `errorMessagesForSchema` function as we strongly believe it is not used by most.
It was created for a necessity of remix-forms library and we'll move it over [to that repo](https://github.com/seasonedcc/remix-forms/issues/219).


diogob (approved):
LGTM

PR #96, opened 2023-08-10
Adds identity and passthrough to the combinators

gustavoguichard:
As discussed in #86 , sometimes it is good to have `identity` function to help on compositions.

I also suggested the `passthrough` which is a shortcut for `map(all(myDf, identity), mergeObjects)` and will probably be used whenever someone wants a domain-function to forward all the received input to the next df in the composition.

I'll keep this PR as a draft until I write proper tests and docs for it.

The plan is to ship it on v2.

### DIFF
--- a/src/domain-functions.ts
+++ b/src/domain-functions.ts
@@ -336,6 +336,21 @@ function trace<D extends DomainFunction = DomainFunction<unknown>>(
   }
 }
 
+// deno-lint-ignore no-explicit-any
+const makeSuccess = (input?: any) =>
+  ({
+    success: true,
+    data: input,
+    inputErrors: [],
+    errors: [],
+    environmentErrors: [],
+    // deno-lint-ignore no-explicit-any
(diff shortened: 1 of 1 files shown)


diogob:
@gustavoguichard I was thinking that our `identity` here is not really an indentity function.

1. It is not a DF identity since it is a DF itself.
2. It is not an identity of the input since it wraps the input in a `SuccessResult`.

So this is really a `SuccessResult` constructor. I suggest we rename it to something that conveys that more clearly.

Some ideas:
- `always`
- `success`
- `makeSuccessResult`

diogob:
@gustavoguichard I took the liberty of reapplying the new passthrough combinator on top of our main. I have also kept the `makeSuccess` (formerly `identity`) private until further deliberation.

diogob:
I'll close this for now since we are still discussing the usefulness of these functions in the library.

PR #97, opened 2023-08-10
feat: Adds a `collectSequence` combinator

gustavoguichard:
For v2 we are adding a `collectSequence` combinator, which has the same API as `collect` but it runs sequentially like `pipe` instead of in parallel.

This feature relies on JS's capability to keep the order of keys of a given object which is true since ECMAScript2015.
For number-like object keys (eg: `{ 2: dfA, 1: dfB }`) JS will follow ascendent order though but we can solve that with docs and warnings.

Still missing docs to merge this PR ;)


gustavoguichard:
This should address #86 .
I hope @jam-fran sees this PR =)

PR #98, opened 2023-08-12
feat: Support other schema validation libs

### LINKED ISSUE #78: Coupling with zod
Hi! I'm curious on the reasoning for picking `zod` as the only validation lib supported. Do you feel that restricts adoption?

I faced a similar situation and ended up building an reusable universal adapter ([`typeschema`](https://typeschema.com)), similar to `tRPC`'s approach. Would something like that be useful to `domain-functions`? I'm open to working together to make this lib available to more devs. Thanks!

decs:
Fixes #78

This PR decouples from `zod` by leveraging `typeschema` and allows using any major schema validation lib (`typia`, `valibot`, `arktype`, etc)

Ran both `deno test` and `deno task build-npm` and all tests passed

As requested, this PR doesn't touch tests or README

### DIFF
--- a/src/branch.test.ts
+++ b/src/branch.test.ts
@@ -3,7 +3,7 @@ import {
   assertEquals,
   assertObjectMatch,
 } from 'https://deno.land/std@0.160.0/testing/asserts.ts'
-import { z } from 'https://deno.land/x/zod@v3.21.4/mod.ts'
+import { z } from 'npm:zod@3.21.4'
 
 import { makeDomainFunction } from './constructor.ts'
 import { branch, pipe, all } from './domain-functions.ts'
--- a/src/collect.test.ts
+++ b/src/collect.test.ts
@@ -3,7 +3,7 @@ import {
   assertEquals,
   assertObjectMatch,
 } from 'https://deno.land/std@0.160.0/testing/asserts.ts'
-import { z } from 'https://deno.land/x/zod@v3.21.4/mod.ts'
+import { z } from 'npm:zod@3.21.4'
 
 import { makeDomainFunction } from './constructor.ts'
 import { collect } from './domain-functions.ts'
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -1,4 +1,4 @@
-import { z } from 'https://deno.land/x/zod@v3.21.4/mod.ts'
+import type { Schema as AnySchema, Infer } from 'npm:@decs/typeschema@0.9.1'
 import type {
   ErrorWithMessage,
   SchemaError,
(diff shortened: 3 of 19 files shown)


gustavoguichard:
Hey @decs , thanks for taking the time! This is interesting.. seems like the dx for zod users wouldn't change.
Could you provide an example of how that'll work for other parsers?

src/branch.test.ts
```diff
@@ -3,7 +3,7 @@ import {
   assertEquals,
   assertObjectMatch,
 } from 'https://deno.land/std@0.160.0/testing/asserts.ts'
-import { z } from 'https://deno.land/x/zod@v3.21.4/mod.ts'
+import { z } from 'npm:zod@3.21.4'
```
diogob, line 6: Why the change from deno.land to npm?

src/collect.test.ts
```diff
@@ -3,7 +3,7 @@ import {
   assertEquals,
   assertObjectMatch,
 } from 'https://deno.land/std@0.160.0/testing/asserts.ts'
-import { z } from 'https://deno.land/x/zod@v3.21.4/mod.ts'
+import { z } from 'npm:zod@3.21.4'
```
diogob, line 6: the fact that we have to change so many imports on a PR like this indicates that we would benefit from some sort of prelude module @gustavoguichard

src/errors.ts
```diff
@@ -1,4 +1,4 @@
-import { z } from 'https://deno.land/x/zod@v3.21.4/mod.ts'
+import type { Schema as AnySchema, Infer } from 'npm:@decs/typeschema@0.9.1'
```
diogob, line 1: I would rather reduce the number of dependencies. Here we are keeping the zod dependency and adding typeschema.

src/branch.test.ts
decs: to be frank, I'm not very familiar with deno, so I might be missing how to do this better. I'd appreciate some advice

typeschema relies on optional peer dependencies to interact with the validation libs. so deno, for some reason, wasn't recognizing `https://deno.land/x/zod@v3.21.4/mod.ts` as being equivalent to `npm:zod@3.21.4` (and that was making the tests fail due to typeschema's `Schema` type not including zod as a valid schema). maybe one way to better fix this is to port typeschema to deno (instead of relying on the npm import), but I'm not sure how to make deno work with optional peer dependencies

src/errors.ts
diogob: I see it now.
We have asked to not change the tests :man_facepalming:

decs:
> Hey @decs , thanks for taking the time! This is interesting.. seems like the dx for zod users wouldn't change. Could you provide an example of how that'll work for other parsers?

It looks the same as how it currently does with zod:

```ts
const schema = type({value: 'number'}); // arktype

// or

const schema = {
  type: 'object',
  properties: {
    value: {type: 'integer'},
  },
  required: ['value'],
  additionalProperties: false,
}; // ajv (JSON schema)

// or

const schema = typeOf<{value: number}>(); // deepkit

// then

const increment = makeDomainFunction(schema)(({ value }) => value + 1)
```

gustavoguichard:
Thank you for the clarifications @decs . I definitely see the value now... 
As soon as I have some time I'll try to play with this PR to give some feedback.
I appreciate the time spent here!

diogob:
I'm moving this to #103

PR #100, opened 2023-08-16
V2

diogob:
##  :fire:  Breaking changes

- Removed the `qs` dependency so some `inputRessolvers` might return different results for a few edge cases. We can elaborate better on the release notes.
- Removed `errorMessagesForSchema`, it seems to be used only in `remix-forms`. We will move this code there.

## :sunglasses: New ferature

- `collectSequence` combinator. It is a sequential version of collect and uses the input object keys' order to determine the sequence of functions to be called.

## :broom:  Housekeeping

- Use a `test-prelude` to keep test imports in a central location, also upgrade our std library version used for testing.


gustavoguichard (approved):
:shipit:

PR #103, opened 2023-08-18
Support other schema validation libs through typeschema

### LINKED ISSUE #78: Coupling with zod

diogob:
See the original #98 for discussion on this feature.

I just moved the development here since we had to resolve some conflicts after rebasing on top of v2.

Edit by @gustavoguichard :
Closes #78


diogob:
@gustavoguichard @decs unfortunatelly I believe we have a breaking change here. It seems that the type-checking errors we had on the tests using the deno version of zod will also happen in any other deno project after this merge.

The runtime works fine, but for some reasons the types are not properly resolved. Perhaps the issue could be solved by tweaking the typeschema npm package, without the need for a deno native module.

I made a small repository as an example: https://github.com/diogob/typeschema-deno-test

diogob:
it seems that for the typeschema Registry to work, the validation library would have to be present in the same modules directory. Therefore, unless the deno project use npm resolvers for the parsers themselves the `Schema` type fails to type-check.

This would force deno users to rewrite all their imports as we had to do for our tests :disappointed:

decs:
Thanks for the investigation! I'll spend some time tomorrow and in the weekend digging on how we can address this. My current plan is to use denoify to generate a Deno port of typeschema that updates all imports to their Deno equivalents. Potentially that should fix the type check.

gustavoguichard:
Hopefully!
It will be beautiful if there's no breaking change at all. If we don't find a way I'd still consider adding it on v3 though.

decs:
Found the root cause! Though the fix isn't that clear yet

typeschema defines the `Schema` type as an union of all the imported types

On Node, if we try importing a module that's not installed, Node doesn't complain, but replaces the type to `any`. To account for that, I added an `IfDefined<T>` helper which converts `any` types to `never` so that the `Schema` union only contains the declared types.

```ts
export type IfDefined<T> = any extends T ? never : T;

export type Schema = {
  [K in keyof Registry]: IfDefined<InferSchema<Registry[K], any>>;
}[keyof Registry];
```

But what I learned is that Deno seems to default to `any` when the union has imported types. As an example:

```ts
import type {BaseSchema} from 'https://deno.land/x/valibot/mod.ts';
import type {ZodSchema} from 'https://deno.land/x/zod/mod.ts';

type Combined = ZodSchema | BaseSchema; // any
```

So the root cause is that, due to this behavior (of converting to `any`), `IfDefined` is converting it later to `never`, which is resulting in these imported types being ignored by the final `Schema` on Deno.

The best approach seems to be just having 2 separate implementations of the `Schema` type: one for Node and another for Deno. But I can't figure out yet on how to create a meaningful type other than `any` for Deno.

decs:
I've ported it to Deno with denoify and it seems to be working now!

@diogob could you try updating this PR to import typeschema with `https://deno.land/x/typeschema@v0.9.3/mod.ts`? then we can preserve the zod imports as they were before

diogob:
@decs it seems we are getting somewhere. If you want to take another stab please feel free to take over this branch on your fork since I won't have much time to investigate the issue.

Another possible way forward would be to go the generic interfaces way and just ensure 2 libraries work out of the box: zod and  typeschema. Doing that we would not have a dependency, we would only have to track some of zod's and tymeschema's types. This should be quite easy since we depend on a few central APIs that should be stable over time.

Moreover we would document that for any library that is not zod, the users can use typeschema as an adaptor and then DF would work out of the box. We could even make both zod and typeschema optional peer deps.

gustavoguichard:
I like this suggestion!

decs:
I couldn't figure out how to take over this PR so my attempt resulted in another PR on top of this branch: #104.
After those tweaks the tests pass and the NPM build works.

Since typeschema is currently based on wrapping functions, we wouldn't be able to rely just on the interface (we need to call `validate()` directly). So DX-wise, having it as a dependency would allow us to make it seamless to end users (i.e. the same as zod today).

One alternative I can implement is updating typeschema to add a `wrap()` function, which would transform any schema in an object with a `.validate()` function. That would allow for using just an interface within DF (i.e. no direct dependency on typeschema), but end users would have to explicitly wrap their schemas:
```ts
// and do
const increment = makeDomainFunction(wrap(schema))(({ value }) => value + 1)
// instead of
const increment = makeDomainFunction(schema)(({ value }) => value + 1)
```

Which do you guys prefer?

diogob:
@decs this PR is looking good. I ask you guys to wait a bit since I'm taking a week off and I don't want to rush this merge. But it seems to be working well and I just want to play a while longer with the code.
If we don't find any other issues I'll be glad to merge as is.

diogob:
Also, thanks again @decs for all the time you have put into this work :sunglasses:

decs:
@diogob I appreciate our collaboration! enjoy your time off!

diogob:
I have update https://github.com/diogob/typeschema-deno-test to test the latest commit in this branch and it works perfectly. Moreover, I tested a local upgrade on an existing project and it worked flawlessly (the only extra dependency added was `typeschema`).

Finally I traced the execution locally and verified that `typeschema`'s `coerce` function ensures that we won't run spurious validators.

The only downside is that in deno projects the `deno.lock` will have the specs of all validator libraries listed as optional dependencies, but since their code will not run on most cases (and deno applicarions use a global cache) I'm willing to pay this price.

Great job @decs

gustavoguichard:
Yeah, I don't mind about the downside either.
Before merging I'll try to test it in an existing project and will let you know;)

gustavoguichard:
Folks, I just tested the changes in a fairly big codebase and there seems to be no breaking changes.
I'm all 👍🏼 for this merge. We just need an update to the README as it is quite focused on Zod. Would you like to give it a try @decs  or @diogob ?

🚢

src/errors.test.ts
```diff
@@ -1,4 +1,5 @@
 import { describe, it, assertEquals } from './test-prelude.ts'
+import { z } from 'https://deno.land/x/zod@v3.21.4/mod.ts'
```
gustavoguichard, line 2:
```suggestion
```

diogob:
I have just converted this back to draft since I might have found some issues using `valibot`. I will create some tests on our side to ensure that our type inference will work with all the supported libs.

decs:
`valibot` had some breaking changes on v0.13.0 and I released v0.10.0 to address that. Could you check if ugrading `typeschema` does the trick?

diogob:
@decs it worked like a charm :tada: 
@gustavoguichard I have added a test in the `constructor.test` to ensure we can create parsers with the latest npm version of valibot. I'm condering adding a test for each library. Thoughts?

gustavoguichard:
I thought the same!
At least test some popular libraries, I wouldn't bother testing all of them bc this is now @decs life mission 🤭

Excited about merging this PR! I really appreciate the initiative and I'm considering adding TypeSchema to [make-service](https://github.com/gustavoguichard/make-service) =D

diogob:
@gustavoguichard I have covered most validators on [TypeSchema coverage](https://github.com/decs/typeschema#coverage) down to valibot.

I wonder if we should create our own list of compatibility, just to steer users away from libraries that have no `Infer` support on TypeSchema. Especially AJV, which seems to be quite popular but is barely usable with DFs since the inference will not work.

Thoughts?

README.md
```diff
@@ -61,7 +61,8 @@ It does this by enforcing the parameters' types at runtime and always wrapping r
 npm i domain-functions zod
 ```
 
-**NOTE :** All our examples will use [zod](https://zod.dev/). You can, however, use any schema validator supported by [TypeSchema](https://typeschema.com/).
+**NOTE :** All our examples will use [zod](https://zod.dev/) which we recommend if you have no existing schema validation library in your project. 
+You can, however, other schema validators supported by [TypeSchema](https://typeschema.com/#coverage).
```
gustavoguichard, line 65: You can, however, USE other

README.md
```diff
@@ -60,6 +61,9 @@ It does this by enforcing the parameters' types at runtime (through [zod](https:
 npm i domain-functions zod
 ```
 
+**NOTE :** All our examples will use [zod](https://zod.dev/) which we recommend if you have no existing schema validation library in your project. 
+You can, however, other schema validators supported by [TypeSchema](https://typeschema.com/#coverage).
```
gustavoguichard, line 65:
```suggestion
You can, however, use other schema validators supported by [TypeSchema](https://typeschema.com/#coverage).
```

gustavoguichard:
@diogob , I like the idea of keeping a list of libs that will get the most of our lib.

gustavoguichard:
@diogob go for it

PR #110, opened 2023-09-12
Expose and add JSDocs to `safeResult`

gustavoguichard:
In order to create our own combinators in our projects to test them before bringing to this repo we will need to expose this function which is used in every combinator.

I added some JSDocs to help whoever decides to hack with domain-functions but not gonna add it to the README yet.


diogob (approved):
LGTM

PR #111, opened 2023-09-12
Export `mdf` alias for `makeDomainFunction`

gustavoguichard:
Inspired by other libs that have shorter versions, the usage of some of our users, and also to make inline compositions less verbose - I am proposing we export the `mdf` alias in this PR.

I think it'd only add a little benefit with no drawback AFAICT.


diogob (approved):
LGTM

PR #112, opened 2023-09-12
Improved `branch` method to act like conditional `pipe`

gustavoguichard:
This PR adds the possibility to return `null` from a `branch` compinator's predicate so it won't pipe and will follow on with the result of the first DF.

This is a non-breaking change to the already existent `branch` as you can see even the types are preserved.
It will only add the type of the first DF to the possible output if the predicate returns null in the conditional return.

The motivation for this PR is that I've tried to come up with other ways to do it with a combination of `branch` and a `passthrough`. But I always ended up breaking the inference with `passthrough` and with this addition the inference is preserved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -656,7 +656,7 @@ For the example above, the result type will be `Result<{ aString: string, aBoole
 
 Use `branch` to add conditional logic to your domain functions' compositions.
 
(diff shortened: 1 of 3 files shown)


diogob (approved):
:fire:

PR #113, opened 2023-09-12
Update README with type inferences and new Remix syntax

gustavoguichard:
Just some updates in the README to reflect new remix APIs and another way to showcase type inference.


diogob (dismissed):
LGTM

PR #114, opened 2023-09-14
Decouple zod

diogob:
Decoupling from zod yields a clearer typing IMHO.
It will also allow anyone to build adapters for any validator as long as they abide by zod's `safeParseAsync` and issues interface.

The long term goal is to track Zod's interface keeping this library always as our de facto standard parser.
The additional burden on type maintenance should be minimal considering these interfaces have a small surface area and are quite fundamental to Zod's backwards compatibility.

## TODO

- [x] Double check if new types are tight enough.
- [x] Write JSDoc for exposed types.
- [x] Write note about adapters in README.


src/constructor.ts
```diff
@@ -118,10 +111,31 @@ function makeDomainFunction<
...
       })
     } as DomainFunction<Awaited<Output>>
   }
 }
 
+// deno-lint-ignore ban-types
+const objectSchema: ParserSchema<{}> = {
+  // deno-lint-ignore require-await
+  safeParseAsync: async (data: unknown) => {
+    if (data == null || typeof data !== 'object') {
```
gustavoguichard, line 124: `typeof [] === "object"` though..
This is the only safe way I know to assure that:
```ts
Object.prototype.toString.call({}) === "[object Object]"
Object.prototype.toString.call([]) === "[object Array]"
Object.prototype.toString.call(null) === "[object Null]"
```

Check out [this function](https://github.com/gustavoguichard/make-service/blob/main/src/internals.ts#L31)

gustavoguichard (commented):
LGTM ❤️

src/constructor.ts
diogob: Yeah, I know the type was wrong (it is in the todo to review these types). But now I see the runtime check is wrong as well!

src/constructor.ts
```diff
@@ -118,10 +111,31 @@ function makeDomainFunction<
...
         }
-        return handler(result.data, envResult.data)
+        return handler(result.data as I, envResult.data as E)
       })
     } as DomainFunction<Awaited<Output>>
   }
 }
 
+// deno-lint-ignore ban-types
+const objectSchema: ParserSchema<{}> = {
```
gustavoguichard, line 121:
```suggestion
const objectSchema: ParserSchema<Record<PropertyKey, unknown>> = {
```

gustavoguichard: In TS `{}` means anything but `null | undefined`

src/constructor.ts
```diff
@@ -118,10 +111,31 @@ function makeDomainFunction<
...
+  // deno-lint-ignore require-await
```
gustavoguichard, line 122: To remove this comment you can remove the `async` and return `Promise.resolve({ success: true, data })`

gustavoguichard (changes requested):
@diogob now I realize that on `objectSchema` and `undefinedSchema`, in case of failure, it should return a `ParserIssue[]` instead of throwing, as they are using a safe method: `safeParseAsync`

And we should also document it ;)

diogob:
> @diogob now I realize that on `objectSchema` and `undefinedSchema`, in case of failure, it should return a `ParserIssue[]` instead of throwing, as they are using a safe method: `safeParseAsync`
> 
> And we should also document it ;)

Great catch! I'm glad we never rush merges :P

gustavoguichard:
I'm getting pretty excited about this PR tho

gustavoguichard (dismissed):

PR #116, opened 2023-10-24
feat: Add async capabilities for the trace function

gustavoguichard:
I'm adding `async` capabilities to the `trace` function because "why not" 😎


diogob (approved):
LGTM

Issue #117, opened 2023-10-29
URIError: URI malformed


ledniy:
Just faced the issue while tried to pass `%` symbol in some of the inputs.

```
error: URIError: URI malformed
      const value = decodeURIComponent(encodedValue)
                    ^
    at decodeURIComponent (<anonymous>)
    at file:///Users/ledniy/Dev/domain-functions/src/input-resolvers.ts:33:21
    at Array.reduce (<anonymous>)
    at inputFromSearch (file:///Users/ledniy/Dev/domain-functions/src/input-resolvers.ts:31:6)
    at inputFromFormData (file:///Users/ledniy/Dev/domain-functions/src/input-resolvers.ts:114:3)
    at Module.inputFromForm (file:///Users/ledniy/Dev/domain-functions/src/input-resolvers.ts:132:3)
    at async Object.<anonymous> (file:///Users/ledniy/Dev/domain-functions/src/input-resolvers.test.ts:68:18)
    at async Function.runTest (https://deno.land/std@0.198.0/testing/_test_suite.ts:358:7)
    at async Function.runTest (https://deno.land/std@0.198.0/testing/_test_suite.ts:346:9)
    at async fn (https://deno.land/std@0.198.0/testing/_test_suite.ts:316:13)
```

Can be reproduced with following test:
```
  it('should parse all symbols correctly', async () => {
    const request = makePost([
      ['formula', '3 % 2']
    ])
    assertEquals(await subject.inputFromForm(request), {
      formula: '3 % 2'
    })
  })
```

gustavoguichard:
Hey @ledniy,
Thank you for reporting!
It doesn't surprise me that it failed but I'll try to investigate it soon and see if we can come up with a fix.

diogob:
@ledniy thanks for the test case

gustavoguichard:
Hey @ledniy , this is now fixed in v2.5.1 .

PR #118, opened 2023-11-01
Fix double decoding URIError when using input resolvers

diogob:
We had an unnecessary `decodeURIComponent` for values passed in forms or search parameters.
The error was introduced during our qs replacement in v2.0.0
This PR should take care of #117

### DIFF
--- a/src/input-resolvers.test.ts
+++ b/src/input-resolvers.test.ts
@@ -23,6 +23,15 @@ const makeGet: (entries: Array<[string, string]>, url?: string) => Request = (
   })
 
 describe('inputFromForm', () => {
+  it('should parse all symbols correctly', async () => {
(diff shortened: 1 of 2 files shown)


src/input-resolvers.test.ts
```diff
@@ -247,7 +256,7 @@ describe('inputFromSearch', () => {
   it('takes keys encoded as URI components', () => {
     const qs = new URLSearchParams()
     qs.append('some%20colors[0]', 'blue')
-    qs.append('some%20colors[1]', 'red%20ish')
```
diogob, line 250: this test case was bogus since `URLSearchParams` will encode values and keys for us.

gustavoguichard (approved):
LGTM

PR #119, opened 2023-11-07
Extracting primitives inspired by ATMP

diogob:
The goal here is to have primitives that don't need parsers and yet compose and handle errors in a similar way as domain functions do. 

Unfortunately this won't provide true type safety on compositions, since these composable primitives will work with any `Composable` function types. 

However there are a few good use cases:

*  To define functions that could act as type-safe interfaces to domain functions by calling a `toComposable` with a fixed type on an already existing DF.
* To create a function with error handling that does not parsers but could potentially be used in DF compositions by calling a `fromComposable` and giving the parsers.
* To bravely compose async functions ~~without any sort of type-safety but~~ without having to juggle awaits and exceptions.

Inspired by the [atmp](https://github.com/gustavoguichard/atmp) library.

This PR is aiming at a "soft launch" where we can release a new version of the lib without breaking changes or any change in documentation.

## TODO

- [x] ensure trace will handle  `traceFn` errors gracefully
- [x] Implement conversion from ATMP -> DF (fromComposable)
- [x] Implement conversion from DF -> ATMP (toComposable)
- [x] Rename ATMP module (renamed to composable)
- [ ] Ensure `Composable` compositions are type-safe (at least when all generics have been passed)


src/domain-functions.ts
```diff
@@ -30,23 +37,10 @@ function all<Fns extends DomainFunction[]>(
...
-            .flat(),
-        })
-      }
-
-      return results.map(({ data }) => data)
-    })
+    const [first, ...rest] = fns.map((df) =>
+      atmp(() => fromSuccess(df)(input, environment)),
+    )
+    return dfResultFromAtmp(A.all(first, ...rest))()
```
gustavoguichard, line 43: Why do we need the destructuring here?
Wouldn't this work?
```suggestion
    const results = fns.map((df) =>
      atmp(() => fromSuccess(df)(input, environment)),
    )
    return dfResultFromAtmp(A.all(results))()
```

diogob: It does not type check. I just assumed that the type was like that by design, but I could be wrong.

gustavoguichard (approved):
LGTM.
I'd add a `cf` alias to `composable` given we already have `mdf` and its success 🤭

diogob:
I'm merging this one since it is backwards compatible and it seems that we are quite comfortable now with the direction it is taking. I'll hold the merge on the primitives type-safety for now.
We also don't need to make any releases for now so we have some change to let ideas percolate with the current main branch code.

PR #120, opened 2023-11-16
upgrade Remix example to Remix_V2

PedroHDP84:
Works like a charm.


diogob (approved):
thanks for the PR @PedroHDP84

PR #121, opened 2023-11-19
Tight composition types

diogob:
This should allow only valid types when using the composition primitives (a.k.a. `Composable` operations).
Since the primitives do not have parsers on their input the type-safety not only makes sense but becomes essential to writing robust compositions.

It is important to note that we rely on a type conversion from a union to a tuple. This type helper (generously shared [here](https://github.com/microsoft/TypeScript/issues/13298#issuecomment-707364842)) uses the internal union order which means that the order when converting from a `Record` to a tuple is not stable. This is irrelevant to type our `collect` function since it is applied in parallel. But I'm leaving this long note here to explain why implementing a `collectSequence` with a Record input would be very challenging.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -5,7 +5,7 @@
     "publish": "deno task build-npm && cd npm/ && npm publish",
     "build-npm": "deno run -A scripts/build-npm.ts",
     "docs": "deno doc --html --name='domain-functions' ./mod.ts"
-  }
+  },
   "lint": {
     "include": ["src/"],
     "rules": {
--- a/deno.json
+++ b/deno.json
@@ -6,4 +6,10 @@
     "build-npm": "deno run -A scripts/build-npm.ts",
     "docs": "deno doc --html --name='domain-functions' ./mod.ts"
   }
+  "lint": {
+    "include": ["src/"],
+    "rules": {
+      "exclude": ["no-explicit-any", "ban-types"]
+    }
+  }
 }
--- a/src/composable/composable.ts
+++ b/src/composable/composable.ts
@@ -94,7 +94,10 @@ type PipeReturn<Fns extends any[]> = Fns extends [
   ? Composable<(...args: P) => O>
   : never
 
-type PipeArguments<Fns extends any[], Arguments extends any[] = []> = Fns extends [
+type PipeArguments<
+  Fns extends any[],
+  Arguments extends any[] = [],
+> = Fns extends [
   Composable<(...a: infer PA) => infer OA>,
   Composable<(...b: infer PB) => infer OB>,
   ...infer rest,
@@ -168,7 +171,10 @@ type SupertypesTuple<
   ? SupertypesTuple<[], restBNoA, [...O, headBNoA]>
   : O
 
-type AllArguments<Fns extends any[], Arguments extends any[] = []> = Fns extends [
+type AllArguments<
+  Fns extends any[],
+  Arguments extends any[] = [],
+> = Fns extends [
   Composable<(...a: infer PA) => infer OA>,
   Composable<(...b: infer PB) => infer OB>,
   ...infer rest,
@@ -193,6 +199,49 @@ type AllArguments<Fns extends any[], Arguments extends any[] = []> = Fns extends
   ? []
   : never
 
+// Thanks to https://github.com/tjjfvi
+// UnionToTuple code lifted from this thread: https://github.com/microsoft/TypeScript/issues/13298#issuecomment-707364842
+// This will not preserve union order but we don't care since this is for Composable paralel application
+type UnionToTuple<T> = (
+  (T extends any ? (t: T) => T : never) extends infer U
+    ? (U extends any ? (u: U) => any : never) extends (v: infer V) => any
+      ? V
+      : never
+    : never
+) extends (_: any) => infer W
+  ? [...UnionToTuple<Exclude<T, W>>, W]
+  : []
+
+type Keys<R extends Record<string, any>> = UnionToTuple<keyof R>
+
+type RecordValuesFromKeysTuple<
+  R extends Record<string, Composable>,
+  K extends unknown[],
+  ValuesTuple extends Composable[] = [],
+> = K extends [infer Head, ...infer rest]
+  ? Head extends string
+    ? rest extends string[]
+      ? RecordValuesFromKeysTuple<R, rest, [...ValuesTuple, R[Head]]>
+      : never
+    : ValuesTuple
+  : ValuesTuple
+
+type Zip<
+  K extends unknown[],
+  V extends Composable[],
+  O extends Record<string, Composable> = {},
+> = K extends [infer HeadK, ...infer restK]
+  ? V extends [infer HeadV, ...infer restV]
+    ? HeadK extends string
+      ? restK extends string[]
+        ? restV extends Composable[]
+          ? Zip<restK, restV, O & { [key in HeadK]: HeadV }>
+          : never
+        : never
+      : never
+    : O
+  : O
+
 /**
  * Receives a Record of Composables, runs them all in parallel and preserves the shape of this record for the data property in successful results.
  * @example
@@ -203,7 +252,9 @@ type AllArguments<Fns extends any[], Arguments extends any[] = []> = Fns extends
  * const df = collect({ a, b })
 //       ^? Composable<() => { a: string, b: number }>
  */
-function collect<T extends Record<string, Composable>>(fns: T) {
+function collect<T extends Record<string, Composable>>(
+  fns: T & Zip<Keys<T>, AllArguments<RecordValuesFromKeysTuple<T, Keys<T>>>>,
+) {
   const [fn, ...fnsWithKey] = Object.entries(fns).map(([key, cf]) =>
     map(cf, (result) => ({ [key]: result })),
   )
--- a/src/composable/composable.ts
+++ b/src/composable/composable.ts
@@ -1,4 +1,4 @@
-import { toErrorWithMessage } from "./errors.ts";
+import { toErrorWithMessage } from './errors.ts'
 import {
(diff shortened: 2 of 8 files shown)


diogob:
@gustavoguichard I think I got `collect` working. Actually the restriction I found only impedes us from typing `collectSequence` since there we need to preserve the order of the keys.

diogob:
@gustavoguichard just realized there is still some work to do unifying types. I have converted back to draft and will mark as ready for review as soon as I finish those bits.

diogob:
Nevermind generalizing the `Unpack` types. I just tried and I remembered why I haven't done it so far. It's not trivial and also not necessary for a release. Once we start gearing up to the major release we can revisit them.

src/composable/types.test.ts
```diff
@@ -7,7 +7,7 @@ export type Equal<A, B> =
   // prettier is removing the parens thus worsening readability
   // prettier-ignore
   (<T>() => T extends A ? 1 : 2) extends (<T>() => T extends B ? 1 : 2) ? true
-    : false
+    : [A, 'should equal', B]
```
gustavoguichard: 🔥

src/composable/composable.ts
```diff
@@ -45,12 +50,12 @@ function composable<T extends Fn>(fn: T): Composable<T> {
   return async (...args) => {
     try {
       // deno-lint-ignore no-explicit-any
-      const result = await fn(...(args as any[]))
-      return success(result)
+      const result = await fn(...(args as any[]));
+      return success(result);
```
gustavoguichard, line 54: what is the deal with the semicolons?

diogob: I clearly spent too much time coding at night :stuck_out_tongue:

diogob:
> > Nevermind generalizing the `Unpack` types.
> 
> You mean being able to use `Unpack` with both `df` and `composable`?

That's what I meant, since these two types have the same name but they behave in quite different ways. The DF `UnpackResult` applies the `ReturnType`, and the composable `UnpackResult` just takes the result out of the promise.
But I wouldn't worry about this for this PR.

src/composable/composable.ts
```diff
@@ -67,15 +72,16 @@ function composable<T extends Fn>(fn: T): Composable<T> {
  * const d = C.pipe(a, b)
  * //    ^? Composable<({ aNumber }: { aNumber: number }) => { aBoolean: boolean }>
  */
-function pipe<T extends [Composable, ...Composable[]]>(...fns: T) {
+function pipe<T extends [Composable, ...Composable[]]>(
+  ...fns: T & PipeArguments<T>
```
diogob, line 76: this trickery is what allows recursive checking of arguments type. `T` is evaluated first and it assumes the inferred type of given arguments, then TS will try to intersect with the result of `PipeArguments`.

deno.json
```diff
@@ -5,5 +5,11 @@
     "publish": "deno task build-npm && cd npm/ && npm publish",
     "build-npm": "deno run -A scripts/build-npm.ts",
     "docs": "deno doc --html --name='domain-functions' ./mod.ts"
+  },
+  "lint": {
+    "include": ["src/"],
+    "rules": {
+      "exclude": ["no-explicit-any", "ban-types"]
```
diogob, line 12: `ban-types` is the lint rule that will forbid `{}` (since one usually want to use a `Record`). This type we could ban in the future but for now I rather reduce the linting noise.

And explicit `any` is very useful in this code-base, so it surprises me we haven't done this sooner :laughing:

src/composable/composable.ts
```diff
@@ -67,15 +71,16 @@ function composable<T extends Fn>(fn: T): Composable<T> {
...
+    //           I don't understand what is the issue here but ignoring the errors
+    //           is safe and much nicer than a bunch of casts to any
+    const res = (await sequence(...fns)(...args))
     return !res.success ? error(res.errors) : success(res.data.at(-1))
-  }) as Composable<
-    (
-      ...args: Parameters<Extract<First<T>, Composable>>
-    ) => UnpackResult<ReturnType<Extract<Last<T>, Composable>>>
-  >
+  }) as PipeReturn<T>
```
diogob, line 83: extracted `PipeReturn` just for the sake of readability

src/composable/composable.ts
```diff
@@ -98,11 +106,7 @@ function all<T extends [Composable, ...Composable[]]>(...fns: T) {
     }
 
     return success((results as Success<any>[]).map(({ data }) => data))
-  }) as unknown as Composable<
-    (...args: Parameters<T[0]>) => {
-      [key in keyof T]: UnpackResult<ReturnType<Extract<T[key], Composable>>>
-    }
-  >
+  }) as AllReturn<T>
```
diogob, line 109: also extracted for readability

src/composable/composable.ts
```diff
@@ -202,8 +209,6 @@ export {
   all,
   collect,
   composable,
-  composable as cf,
-  composable as λ,
```
diogob, line 206: I think `composable` is short enough and the lack of aliases creates a nice uniformity to code-bases. I'm not strongly opposed to reintroducing them however.

src/domain-functions.ts
```diff
@@ -70,10 +71,10 @@ function all<Fns extends DomainFunction[]>(
   ...fns: Fns
 ): DomainFunction<UnpackAll<Fns>> {
   return ((input, environment) => {
-    const [first, ...rest] = fns.map((df) =>
-      A.λ(() => fromSuccess(df)(input, environment)),
+    const composables = fns.map((df) =>
+      A.composable(() => fromSuccess(df)(input, environment)),
     )
-    return dfResultFromcomposable(A.all(first, ...rest))()
+    return dfResultFromcomposable(A.all(...(composables as [Composable])))()
```
diogob, line 77: we need the typecast since `A.all` will not take empty arrays and the map will always resolve to a type that is potentially empty.

src/domain-functions.ts
```diff
@@ -212,10 +213,12 @@ function sequence<Fns extends DomainFunction[]>(
...
 ): DomainFunction<UnpackAll<Fns>> {
   return function (input: unknown, environment?: unknown) {
-    const [first, ...rest] = fns.map((df) =>
-      A.λ(fromSuccess(applyEnvironment(df, environment))),
+    const dfsAsComposable = fns.map((df) =>
+      A.composable(fromSuccess(applyEnvironment(df, environment))),
     )
-    return dfResultFromcomposable(A.sequence(first, ...rest))(input)
+    return dfResultFromcomposable(
+      A.sequence(...(dfsAsComposable as [Composable])),
```
diogob, line 220: I didn't fully understand why this type cast had to be written as a one element tuple since `sequence` will take any non-empty list. But this type cast should have no effect outside the function so I'm happy to leave it as is.

src/composable/composable.ts
```diff
@@ -98,8 +105,8 @@ function all<T extends [Composable, ...Composable[]]>(...fns: T) {
     }
 
     return success((results as Success<any>[]).map(({ data }) => data))
-  }) as unknown as Composable<
-    (...args: Parameters<T[0]>) => {
+  }) as Composable<
```
diogob, line 108: I have tried to extract `AllReturn` but this will prevent TS from showing us the full inline resolved type yielding a poorer DX. We can do this in the future in case we find out how to tell TS to inline resolved types.

src/composable/composable.ts
```diff
@@ -115,12 +122,14 @@ function all<T extends [Composable, ...Composable[]]>(...fns: T) {
...
-  const [fn, ...fnsWithKey] = Object.entries(fns).map(([key, cf]) =>
+function collect<T extends Record<string, Composable>>(
+  fns: T & CollectArguments<T>,
+) {
+  const fnsWithKey = Object.entries(fns).map(([key, cf]) =>
     map(cf, (result) => ({ [key]: result })),
   )
-  return map(all(fn, ...fnsWithKey), mergeObjects) as Composable<
-    (...args: Parameters<Extract<T[keyof T], Composable>>) => {
+  return map(all(...(fnsWithKey as any)), mergeObjects) as Composable<
```
diogob, line 131: Same issue as `AllReturn` prevented me from extracting this one.

src/composable/composable.ts
gustavoguichard: Ok, this explanation is what clicked for me! TY

deno.json
gustavoguichard: I don't like that linting rule because I like to use `{}` as it is equivalent to something like `Omit<any, undefined | null>`

src/domain-functions.ts
gustavoguichard: to make it like a not empty array instead of a tuple of size 1 you can use `[Composable, ...Composable[]]`

src/domain-functions.ts
gustavoguichard: Check if my previous comment applies here

gustavoguichard (approved):
This PR is intimidating but I believe I'll catch up in the next few iterations.
As the types get more complex I start to feel the need to properly name the generic params or at least we should have some dictionary of standard letters so we know things like:
- `T` is the first generic argument for simple types
- `U` is the second argument
- `I` and `O` means Input and Output
- etc...

I've been writing lowercased full words for more complex types in string-ts and kept the `T` and `U` in some places.

</rambling>

src/domain-functions.ts
diogob: That was my first intuition as well, but the `AllArguments` bit messes up this case. Perhaps we still need to fix something for cases where I want to pass a list of functions, but since this is not the main use case for compositions (and we are not documenting this yet) I'm happy to merge as is.

diogob:
I have a proposal for new naming conventions, will send in another PR to avoid muddying the waters here

PR #125, opened 2023-11-29
Do not error when passing parameters to a DF created with makeDomainFunction()

diogob:
Based on [this discussion](https://github.com/seasonedcc/domain-functions/discussions/122)
Note that this is a breaking change, even though unlikely, someone might be relying on the error of an undefined parser.

I think it is a nicer default behaviour though. And the previous behaviour is easy to reproduce by passing a `z.undefined()`. We can add this note on the release.

### DIFF
--- a/src/constructor.test.ts
+++ b/src/constructor.test.ts
@@ -68,13 +68,14 @@ describe('makeDomainFunction', () => {
     })
 
     it('fails gracefully if gets something other than undefined', async () => {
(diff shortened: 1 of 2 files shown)


gustavoguichard (approved):
LGTM

diogob:
I'll hold this merge since we might want to release a minor version with the primitives before 3.0

diogob:
This will be part of the next big breaking release. They are composable :sunglasses:

PR #127, opened 2024-01-05
New sequenceAll combinator

diogob:
## Use case

This fits a similar use case as the passthrough proposed on #96
It can be used to add validation functions to a pipeline that should abort execution on error but don't necessarly produce a result that we want to pipe to the next callee.

For instance, let's say you want to check the state of a record in the database and send a notification only when the validation function is successful.
In this case the resulting type of the validation function should not be taken into account, since you don't want to write the function for the composition.

With the new combinator something like this could be done:
```typescript
const isValidInDatabase = composable(id: string) => { ... }
const sendNotification = composable(id: string) => { ... }
const sendNotificationWhenValid = sequenceAll(isValidInDatabase, sendNotification)
```

It also makes it easier to reuse existing functions just for the sake of their side-effects. On the previous example, let's say you want add a function that records a log of messages sent on the database (and fail to send the message in case the log fails).

I was hoping to merge this before we have the full composable documentation.

## TODO

- [ ] Add DomainFunction version of the combinator
- [ ] Add README documentation for DomainFunction.

### DIFF
--- a/src/composable/composable.ts
+++ b/src/composable/composable.ts
@@ -140,8 +140,8 @@ function collect<T extends Record<string, Composable>>(
  * @example
  * import { cf as C } from 'domain-functions'
  *
- * const a = C.compose((aNumber: number) => String(aNumber))
- * const b = C.compose((aString: string) => aString === '1')
+ * const a = C.composable((aNumber: number) => String(aNumber))
+ * const b = C.composable((aString: string) => aString === '1')
  * const cf = C.sequence(a, b)
  * //    ^? Composable<(aNumber: number) => [string, boolean]>
  */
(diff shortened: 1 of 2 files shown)


gustavoguichard:
@diogob , I'm good with having it but the name doesn't do it for me... hard to tell what it does, especially because we have `sequence` and `all` and it has nothing to do with them. WDYT?

src/composable/composable.ts
```diff
@@ -166,6 +166,40 @@ function sequence<T extends [Composable, ...Composable[]]>(
   >
 }
 
+/**
+ * Sequences a tuple of functions and collect the output of every function like `sequence` feeding the same input to all of them.
+ * Will abort the sequence on the first error.
+ * @example
+ * import { composable as C } from 'domain-functions'
+ *
+ * const a = C.composable((something: unknown) => String(something))
```
gustavoguichard, line 175: Maybe change the output type to better show the input of the next function is not related to it

```suggestion
 * const a = C.composable((something: unknown) => Number(something))
```

diogob:
> @diogob , I'm good with having it but the name doesn't do it for me... hard to tell what it does, especially because we have `sequence` and `all` and it has nothing to do with them. WDYT?

I actually derived the name from both, since it takes the typing of `all` but most of the implementation from `sequence`. You could say that this is a serial version of the `all` combinator.

Other possible names along this lines would be:

* `allInSequence`
* `allInSeries`
* `serialAll`

gustavoguichard:
tap?

PR #129, opened 2024-01-22
Fix composable pipe for async functions

diogob:
The types `PipeArguments` and `PipeReturn` were not taking into account async composables (the runtime already had the proper awaits though).

The fix using `Awaited` caused functions returning never to type check, so I introduced the `IsNever` and handled those cases separately.


gustavoguichard (approved):
LGTM

PR #130, opened 2024-02-09
Fix mapError exception handling

diogob:
We should not allow exceptions to bubble up freely in composables since this could ruin our composition logic.
I have also added a test case for `map` since we had none.


gustavoguichard (approved):
LGTM

PR #132, opened 2024-02-23
Simplify error type

diogob:
This PR brings a breaking change proposed for the next major version.
It drops the optional `exception` property present in `ErrorWithMessage` in favour of always using an `Error` type to represent errors.

The `Error` already carries a message, hence once the exception becomes mandatory the wrapping object becomes entirely redundant.
Moreover, having an optional `exception` does not make a lot of sense since the only way to return a result with the `errors` property is by using exceptions.

The code-base gets a bit tidier and checking for exception types becomes simpler. There is also didactic value since every TS developer should already be familiar with `Error`.

- Ensure our ErrorWithMessage type always has an attched exception
- Simplify ErrorWithMessage so it's just an Error alias

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ The error result has the following structure:
 ```ts
 type ErrorResult = {
   success: false
-  errors: ErrorWithMessage[]
+  errors: Error[]
   inputErrors: SchemaError[]
   environmentErrors: SchemaError[]
 }
--- a/src/all.test.ts
+++ b/src/all.test.ts
@@ -76,7 +76,7 @@ describe('all', () => {
 
     assertEquals(await c({ id: 1 }), {
       success: false,
-      errors: [{ message: 'Error', exception: 'Error' }],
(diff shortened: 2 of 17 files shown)


gustavoguichard:
I need to think through it. My main concern is how easy is it to work with at the UI level, are we gonna have helpers to extract the messages? Also, how does it affect InputErrors, EnvironmentErrors, and other abstractions people may be using.

diogob:
`InputErrors` and `EnvironmentErrors` will no be affected as far as I can tell, since they deal with `SchemaError`.

diogob:
A smoother approach would be to make exception mandatory and create a function to access the exception (and perhaps the message). This would make for a tighter abstraction and easier future changes.

gustavoguichard:
Ok, I think I like it, I've been dealing with `errors[i].exception` a lot lately anyway

diogob:
composable-functions is upon us!

PR #133, opened 2024-03-01
catchError combinator to recover from failed composable

diogob:
Implement combinator based on the [docs for a future version](https://github.com/seasonedcc/domain-functions/blob/composable-docs/composables.md).

I have changed its type since the document was written since it seems to me that the original input must be relayed for proper error handling (e.g. retry policies).

### DIFF
--- a/src/composable/composable.ts
+++ b/src/composable/composable.ts
@@ -185,6 +185,30 @@ function map<T extends Composable, R>(
 }
 
 /**
+ * Creates a new function that will try to recover from a resulting Failure. When the given function succeeds, its result is returned without changes.
+ * @example
+ * import { cf as C } from 'domain-functions'
+ *
+ * const increment = C.composable(({ id }: { id: number }) => id + 1)
+ * const negativeOnError = C.catchError(increment, (result, originalInput) => (
+ *  originalInput.id * -1
+ * ))
+ */
+function catchError<T extends Composable, R>(
+  fn: T,
+  catcher: (
+    err: Omit<Failure, 'success'>,
+    ...originalInput: Parameters<T>
+  ) => UnpackResult<ReturnType<T>>,
+) {
+  return (async (...args) => {
+    const res = await fn(...args)
+    if (res.success) return success(res.data)
+    return composable(catcher)(res, ...(args as any))
+  }) as T
+}
+
+/**
  * Creates a new function that will apply a transformation over a resulting Failure from the given function. When the given function succeeds, its result is returned without changes.
  * @example
  * import { cf as C } from 'domain-functions'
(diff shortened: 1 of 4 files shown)


src/domain-functions.ts
```diff
@@ -168,7 +168,7 @@ function pipe<T extends DomainFunction[]>(
   ...fns: T
 ): DomainFunction<Last<UnpackAll<T>>> {
   const last = <T>(ls: T[]): T => ls[ls.length - 1]
-  return map(sequence(...fns), last)
+  return map(sequence(...fns), last) as DomainFunction<Last<UnpackAll<T>>>
```
gustavoguichard, line 171: What happened here?

src/composable/composable.ts
```diff
@@ -184,6 +184,38 @@ function map<T extends Composable, R>(
...
+  fn: Composable<T>,
+  catcher: (
+    err: Omit<Failure, 'success'>,
+    ...originalInput: Parameters<T>
+  ) => R,
+) {
+  return (async (...args: Parameters<T>) => {
+    const res = await fn(...args)
+    if (res.success) return success(res.data)
+    return composable(catcher)(res, ...(args as any))
```
gustavoguichard, line 207: Just to make sure you chose it consciously: the first argument to catcher is an actual ErrorResult, not just the errors, right?

Maybe it would be cool to add a test case using the error on the catcher just to feel how the API feels like.

gustavoguichard (approved):

src/domain-functions.ts
diogob: this is unrelated but the tests would break in the latest deno version without this change.
They have [upgraded TS to 5.4.3](https://deno.com/blog/v1.42) and it changed the way TS resolves the type here.

After the upgrade I get
```
error: TS2322 [ERROR]: Type 'DomainFunction<unknown>' is not assignable to type 'DomainFunction<Last<UnpackAll<T>>>'.
  Type 'unknown' is not assignable to type 'Last<UnpackAll<T>>'.
  return map(sequence(...fns), last)
  ~~~~~~
    at file:///home/diogo/Projects/seasoned/domain-functions/src/domain-functions.ts:171:3
```

It resolves `map(sequence(...fns), last)` to `DomainFunction<unknown>`, which is true since we are applying `fns` which is something that extends `DomainFunction<unknown>`. Now that I think about it, it surprises me that it was type-checking before.

I can extract this to a separate PR if you think it's too bad to leave it here.

src/composable/composable.ts
diogob: That was an oversight, the API is certainly more elegant receiving errors.

diogob: I have tweaked an existing test case just to properly use the entire API.

gustavoguichard (approved):

PR #134, opened 2024-03-08
Allow map and mapError in DFs to have async mapper function

diogob:
Do we have any reason to not allow this?


gustavoguichard (approved):
No reason not to

PR #135, opened 2024-03-10
Reimplement composable map in terms of pipe

diogob:
This implementation seems cleaner and easier to understand.

Also simplified the DF map since I thought that the current code was still a bit awkward passing and applying parameters.


gustavoguichard (approved):
👨‍🍳💋

PR #138, opened 2024-04-08
feat: Add and export makeSuccessResult and makeErrorResult helpers

gustavoguichard:
Often times, especially in tests, we want to construct the `Result` type and we have to type all the structure.
These functions should help with that by allowing us to construct an ErrorResult like so:
```ts
// before: 
const error: ErrorResult = {
  success: false,
  errors: [myError],
  environmentErrors: [],
  inputErrors: [],
}

// after:
const error: ErrorResult = makeErrorResult({ errors: [MyError] })
```

or a SuccessResult like so:
```ts
// before: 
const result: SuccessResult = {
  success: true,
  data,
  errors: [],
  inputErrors: [],
  environmentErrors: [],
}

// after:
const result: SuccessResult = makeSuccessResult(data)
```

### DIFF
--- a/src/all.test.ts
+++ b/src/all.test.ts
@@ -22,7 +22,7 @@ describe('all', () => {
 
     assertEquals(
       await c({ id: 1 }),
-      makeSuccessResult([2, 0] as [number, number]),
+      makeSuccessResult<[number, number]>([2, 0]),
     )
   })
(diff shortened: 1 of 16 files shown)


diogob (approved):
LGTM

PR #139, opened 2024-05-22
Reimplement Domain functions with code from the future

diogob:
An alternate timeline


src/constructor.ts
```diff
@@ -4,9 +4,142 @@ import {
...
+  return (async (input?: unknown, environment?: unknown) => {
+    const result = await df(input, environment)
+
+    if (result.success) {
+      return {
+        success: true,
+        data: result.data,
+        errors: [],
+        inputErrors: [],
+        environmentErrors: [],
```
gustavoguichard, line 47:
```suggestion
```

README.md
```diff
@@ -1,4 +1,10 @@
-# Keep your business logic clean with Domain Functions
+# Keep your business logic clean with ~Domain Functions~ Composable Functions
+
+The work in Domain Functions led us to a new, more generic library called Composable Functions.
+We recommend you migrate following [our migration guide](https://github.com/seasonedcc/composable-functions/blob/main/migrating-df.md).
+This guide also has some benefits that we see in migrating to the new library.
+Moreover, to make the transition smoother, there is a new Domain Functions release (3.0) that introduces very few breaking changes and
+uses the new library as a dependency. Using both libraries simultaneously is easy and an incremental migration is viable.
```
gustavoguichard, line 7:
```suggestion
> The work in Domain Functions led us to a new, more generic library called Composable Functions.
We recommend you migrate following [our migration guide](https://github.com/seasonedcc/composable-functions/blob/main/migrating-df.md). `domain-functions` is going into a feature freeze  and we'll be only be doing minimal maintenance to solve crucial bugs.
> This guide also has some benefits that we see in migrating to the new library.
> Moreover, to make the transition smoother, there is a new Domain Functions release (3.0) that introduces very few breaking changes and uses the new library as a dependency. Using both libraries simultaneously is easy and an incremental migration is viable.
```

PR #142, opened 2024-05-24
Add a function for sequential compositions with same arguments

gustavoguichard:
TODO:
- [ ] Name it properly
- [ ] JSDocs
- [ ] Docs on API reference
- [ ] Release and Ping #86

### DIFF
--- a/src/combinators.ts
+++ b/src/combinators.ts
@@ -166,6 +166,27 @@ function sequence<Fns extends [Composable, ...Composable[]]>(...fns: Fns) {
 }
 
 /**
+ * TODO
+ * @param fns TODO
+ * @returns TODO
+ */
+function tap<Fns extends Composable[]>(...fns: Fns) {
+  return (async (...args) => {
+    const result = []
+    for await (const fn of fns) {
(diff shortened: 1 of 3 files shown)


gustavoguichard:
I'm not sure how to name this function. `tap` seems like a nice name but it is not quite what is happening here.
It is probably one of the applications of this combinator, like running side effects in sequence.

Ramda has a definition for tap: https://ramdajs.com/0.21.0/docs/#tap
Rxjx also has a definition for it: https://rxjs.dev/api/index/function/tap

diogob:
> Ramda has a definition for tap: https://ramdajs.com/0.21.0/docs/#tap 
> Rxjx also has a definition for it: https://rxjs.dev/api/index/function/tap

I think what bothers me with the `tap` name is that on both the examples it refers to some subject (`object` in Ramda and `Observable` in RxJS) we return once the computations as done.

In our case, since we have only a function that receives all the computations and must return another computation, the name `tap` feels counter-intuitive.

gustavoguichard:
Just to register it here too: What about `allInSequence`?

PR #143, opened 2024-05-29
Explicit return types

diogob:
Tries to make our typing more elegant and safe.
Also, it should theoretically improve performance (unverified claim based on deno guides)

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -50,7 +50,7 @@ type Composable<T extends (...args: any[]) => any = (...args: any[]) => any> = (
 ) => Promise<Result<Awaited<ReturnType<T>>>>
 
 /**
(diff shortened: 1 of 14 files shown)


src/types.ts
```diff
@@ -49,6 +49,13 @@ type Composable<T extends (...args: any[]) => any = (...args: any[]) => any> = (
   ...args: Parameters<T>
 ) => Promise<Result<Awaited<ReturnType<T>>>>
 
+/**
+ * A composable async function that catches failures.
```
gustavoguichard:
```suggestion
 * A composable async function with schema validation at runtime.
```

gustavoguichard (approved):

PR #145, opened 2024-05-30
Invert parameters in apply-schema so we can validate schema type based on composable parameters

diogob:
rough draft, still need to update examples and docs

not so sure about this one though, perhaps it is best to keep the current solution. I'm leaving the PR open so we can contemplate de diff a while longer.


diogob:
we might reopen this in the future

Issue #146, opened 2024-06-05
TS won't complain about implicit any in composable declaration


danielweinmann:
Example:

![image](https://github.com/seasonedcc/composable-functions/assets/204765/046770d2-fb1e-48db-9589-79c947e18db4)

gustavoguichard:
I don't get this bug, isn't it what [TS does by default](https://www.typescriptlang.org/play/?target=6#code/IYZwngdgxgBAZgV2gFwJYHsI2QJ2FAawCF1gcATAGVPIFNyAKAKBlZgG8YAHHdAK1pRkASXIwAvgBoWbAPSyYAPQD8M1rQgA3VLwgBbDcmlsY8paoCUHGTlrIEOLBAQAbF03FA)?

![Screenshot 2024-06-05 at 13 01 57](https://github.com/seasonedcc/composable-functions/assets/566971/caca8fb2-8f7a-4146-8bf2-38a3816a2ff2)

What are we aiming for here?
IMO you should type your parameters as you do with any function, but wrapping it in a `composable` will give ou composability and monadic error handling. When using `withSchema` on the other hand, the parameters will infer the type of the schema given earlier and if none is given then the parameters will be `unknown`.

danielweinmann:
I want to have the same error I get when using strict TS inside of composable. Here's an example:

```ts
const foo = (bar) => bar
// ^ Error TS7006: Parameter bar implicitly has an any type.

const foo2 = composable((bar) => bar)
// ^ No error whatsoever
```

gustavoguichard:
Fair enough... so the problem is not that it is `any` but the fact the lint isn't complaining, right?
I'll try to replicate to see what I can do about it

diogob:
@gustavoguichard to be more precise, that is a TS error (no linting required).

PR #147, opened 2024-06-07
No implicit any

gustavoguichard:
Solves #146 

This seemed to be impossible, I've done a lot of research and created [an X thread](https://x.com/gugaguichard/status/1798802769996136459) with some experts.

The solutions seemed to be:
- Forbid `any` at all which seems reasonable
- Swap `(...args: any[]) => any` for `(...args: never[]) => unknown`. This one was suggested by our contributor @jly36963 which is a thread on [TS repo itself](https://github.com/microsoft/TypeScript/issues/37595#issuecomment-604103234), and it seems nice because if you don't declare the types the arguments are gonna be never so you can't use them. I was a bit worried about some comments about problems with default arguments - as you can read in the last comment on that thread.

This morning I decided to try and use the `Function` primitive. I know we are not supposed to use it for several reasons (it could mean any constructor, or callable object, etc) but it seems to do it here.

WDYT?

### DIFF
--- a/src/tests/constructors.test.ts
+++ b/src/tests/constructors.test.ts
@@ -151,7 +151,7 @@ describe('withSchema', () => {
       assertEquals(await handler(), success('no input!'))
     })
 
-    it('ignores the input and pass undefined', async () => {
+    it('defaults non-declared input to unknown', async () => {
       const handler = withSchema()((args) => args)
       type _R = Expect<
         Equal<
--- a/src/tests/constructors.test.ts
+++ b/src/tests/constructors.test.ts
@@ -49,6 +49,12 @@ describe('composable', () => {
     assertEquals(res, success(3))
   })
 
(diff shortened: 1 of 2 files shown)


src/tests/constructors.test.ts
```diff
@@ -110,7 +116,7 @@ describe('fromSuccess', () => {
     const a = composable(() => 1)
 
     const c = fromSuccess(a)
-    type _R = Expect<Equal<typeof c, () => Promise<number>>>
+    type _R = Expect<Equal<typeof c, () => Promise<1>>>
```
gustavoguichard, line 119: @diogob I don't know why it changed.. narrower seems better. Didn't break any other test... any ideas?

diogob: I don't understand, but LGTM

diogob (approved):
looks awesome

PR #148, opened 2024-06-07
apply schema

diogob:
We should not allow `applySchema` to take schemas that would potentially break our runtime with a wider type than what the composable takes.


gustavoguichard (commented):
I'll need to update the arktype example before merging it

gustavoguichard (approved):

PR #152, opened 2024-06-19
Rename from `environment` to `context`

gustavoguichard:
Environment has always been a word which required some "context" (🥁) to be understood while `context` is arguably easier.

This change won't be breaking because we are deprecating `EnvironmentError` and `isEnvironmentError` and it can be easily replaced by `ContextError` and `isContextError`.


src/errors.ts
```diff
@@ -29,13 +30,34 @@ class InputError extends Error {
...
   /**
-   * Path of environment attribute that originated the error.
+   * Path of context attribute that originated the error.
    */
   path: string[]
 
   constructor(message: string, path: string[] = []) {
     super(message)
-    this.name = 'EnvironmentError'
+    this.name = 'ContextError'
```
gustavoguichard: Maybe this one should still be:
```suggestion
    this.name = 'EnvironmentError'
```

Because we are checking for it on both `isEnvironmentError` and `isContextError`

src/tests/errors.test.ts
```diff
@@ -10,14 +12,23 @@ describe('isInputError', () => {
...
 })
 
-describe('isEnvironmentError', () => {
-  it('checks if an error is instance of EnvironmentError', () => {
+describe('isContextError', () => {
+  it('checks if an error is instance of ContextError', () => {
+    assertEquals(isContextError(new Error('No')), false)
+    assertEquals(isContextError(new InputError('No')), false)
+    assertEquals(isContextError(new ContextError('Yes')), true)
+    assertEquals(isContextError(new EnvironmentError('Yes')), true)
```
gustavoguichard, line 25: These are important in this change

src/tests/errors.test.ts
```diff
@@ -10,14 +12,23 @@ describe('isInputError', () => {
...
+  })
+
+  it('checks if an error is instance of legacy EnvironmentError', () => {
     assertEquals(isEnvironmentError(new Error('No')), false)
     assertEquals(isEnvironmentError(new InputError('No')), false)
+    assertEquals(isEnvironmentError(new ContextError('Yes')), true)
```
gustavoguichard, line 31: So are these 2 tests

src/errors.ts
diogob: I think it might indeed be easier to keep the previous `EnvironmentError` name until we release the breaking change, just in case there is someone expecting this name on the serialized data and accessing it directly.

diogob (approved):
LGTM

PR #153, opened 2024-06-21
Build a composable that always succeeds

diogob:
I have written this function more than a couple of times and a see some use cases in the wild but always brush it off thinking "why bother, it is so simple to reimplement".

The same argument can be used to bring it in though. It's very low complexity and it helps to keep the typing consistent.

### DIFF
--- a/src/tests/constructors.test.ts
+++ b/src/tests/constructors.test.ts
@@ -13,12 +13,14 @@ import type {
   Success,
 } from '../index.ts'
 import {
+  always,
   composable,
   ContextError,
   ErrorList,
   failure,
   fromSuccess,
   InputError,
+  pipe,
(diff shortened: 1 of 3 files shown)


src/tests/constructors.test.ts
```diff
@@ -431,3 +433,46 @@ describe('applySchema', () => {
...
+  })
+
+  it('allows composition with any sort of input', async () => {
+    const maySucceed = pipe(
+      always({ willBeDiscarded: true }),
+      always('any data'),
+    )
+
+    // the pipe introduces the possibility of failure
+    // our typing is not yet smart enough to detect that all Results are success
```
gustavoguichard, line 470: TODO

PR #154, opened 2024-06-22
Accept plain functions in combinators

gustavoguichard:
This is a work in progress for a POC.
The idea is to be able to accept any function in combinators, so we simplify their usage:
```ts
const add = (a: number, b: number) => a + b

// from:
const fn = collect({ add: composable(add), toString: map(composable(add), String) })

// to:
const fn = collect({ add, toString: map(add, String) })
```

TODO:
- [ ] Fix the BranchReturn type
- [ ] Fix some types responsible for merging parameters of functions - when testing collect with a composable and a plain function the parameters resolved to never.

### DIFF
--- a/src/combinators.ts
+++ b/src/combinators.ts
@@ -13,7 +13,16 @@ import type {
   Success,
   UnpackData,
 } from './types.ts'
-import { composable, failure, fromSuccess, success } from './constructors.ts'
+import {
+  composable,
+  ensureComposable,
+  failure,
+  fromSuccess,
+  success,
+} from './constructors.ts'
+import { EnsureComposable } from './constructors.ts'
+
+type AnyFn = (...args: any[]) => any
 
 /**
  * Merges a list of objects into a single object.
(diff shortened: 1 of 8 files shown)


diogob:
Interesting idea.

src/combinators.ts
```diff
@@ -270,12 +281,13 @@ function catchFailure<
   (
     ...args: Parameters<Fn>
   ) => Awaited<ReturnType<C>> extends never[]
-    ? UnpackData<Fn> extends any[] ? UnpackData<Fn>
-    : Awaited<ReturnType<C>> | UnpackData<Fn>
+    ? UnpackData<Fn> extends any[]
```
diogob, line 284: this looks like prettier, lets keep deno fmt?

gustavoguichard: Yeah, this is far from done. I'm gonna clean it up once it starts to get serious ;)

gustavoguichard:
Closing it in favor of #155

PR #155, opened 2024-06-24
Accept Plain functions in every combinator

gustavoguichard:
This PR introducing the tagging of composables at both type and runtime levels: `fn.kind = 'composable'`

This allows us to achieve: `Composable<Composable<() => void>> === Composable<() => void>`.

And by doing so we are now able to accept plain functions in any combinator, like so:

```ts
const add = (a: number, b: number) => a + b

// Before this PR:
const fn = collect({ add: composable(add), toString: map(composable(add), String) })

// With this PR:
const fn = collect({ add, toString: map(add, String) })
```

There are still 2 todos to my goal:
- [x] Make the `branch` resolver function be able to return plain functions instead of only composables or null
- [x] Make the `applySchema` work with plain functions and then we can deprecate `withSchema` because they will be virtually the same

I think this PR as is can be merged and the remaining tasks can come later as new features.


diogob (dismissed):
:sunglasses:

PR #158, opened 2024-06-27
Documentation for new Plain-functions feature

gustavoguichard:


diogob (approved):
LGTM

README.md
```diff
@@ -150,15 +160,16 @@ const addAndReturnString = pipe(toString, add)
 Since pipe will compose from left to right, the only `string` output from `toString` will not fit into the first argument of `add` which is a `number`.
 The error message comes in the form of an inferred `FailToCompose` type. This failure type is not callable, therefore it will break any attempts to call `addAndReturnString`.
 
-### Using non-composables (mapping)
+### Transforming the output (mapping)
 
-Sometimes we want to use a simple function in this sort of sequential composition. Imagine that `toString` is not a composable, and you just want to apply a plain old function to the result of `add` when it succeeds.
-The function `map` can be used for this, since we are mapping over the result of a `Composable`:
+Sometimes we want to use a simple function to transform the output of another function. Imagine you want to apply a plain old function to the result of `add` when it succeeds.
+The function `map` can be used for this:
```
diogob, line 166: I guess we will eventually have assign `map` a different _raison d'être_ since there are several cases now where you could just use pipe.

gustavoguichard: Yes, thought about it yesterday but here follows the differences:
- pipe accepts many functions
- map passes the input to the second function
- map infers the arguments of second function while you've got to declare them in pipe


Having thought about that I believe there are places for both to coexist

diogob: I do believe there is, and I find map is probably going to be a very convenient way to transform output. We have, however, to find a good concept to encapsulate the differences you have pointed out to use in the docs.

PR #159, opened 2024-06-27
Accept plain functions in `branch` Resolver

gustavoguichard:
With some tackling I was able to make it work and now we are ready for v4.2.

It solves one of the remaining todos of #155


diogob (approved):
:100:

Issue #163, opened 2024-07-15
applySchema does not accept plain functions


danielweinmann:
I get the following error when trying to use `applySchema` with a plain function:

```ts
  const increment = (value: number) => value + 1
  const incrementWithSchema = applySchema(z.number())(increment)

// Argument of type '(value: number) => number' is not assignable to parameter of type '((input: number, context: unknown) => Promise<Result<unknown>>) & { kind: "composable"; }'.
//   Type '(value: number) => number' is not assignable to type '(input: number, context: unknown) => Promise<Result<unknown>>'.
//     Type 'number' is not assignable to type 'Promise<Result<unknown>>'.ts(2345)
```

If I wrap it with `composable`, everything works fine.

```ts
  const increment = composable((value: number) => value + 1)
  const incrementWithSchema = applySchema(z.number())(increment)
```

gustavoguichard:
Yup, this is planned. And, once we do that, `applySchema` and `withSchema` will do the same thing. Use withSchema for plain functions in the meantime

gustavoguichard:
@danielweinmann #164 should solve this issue =)

gustavoguichard:
Solved on v4.3

PR #164, opened 2024-07-15
Accept plain functions in `applySchema` and deprecates `withSchema`

gustavoguichard:
With these changes I was able to replace every occurrence of  `withSchema` with `applySchema` and the tests are passing.
Now `applySchema` accepts plain functions and infers the types of the function's parameters.

I also deprecated `withSchema` and updated docs and examples to omit its usage.

This should solve #163 

It also closes the latest remaining task of #155 =D


diogob (approved):
LGTM

PR #170, opened 2025-02-05
Standard Schema Spec

gustavoguichard:
This PR will allow us to work with a bunch of schema libraries, such as Zod, Arktype, Valibot, and others.
Check the [Standard Schema Spec](https://standardschema.dev) for more information.

This PR is probably breaking and we should prepare a new major version of composable-functions.
I'll leave it open so we can discuss it @diogob


src/constructors.ts
```diff
@@ -110,28 +110,34 @@ function fromSuccess<O, P extends any[]>(
...
+        .validate(
+          context,
+        )
+      const result = (inputSchema ?? alwaysUnknownSchema)['~standard'].validate(
+        input,
       )
-      const result = (inputSchema ?? alwaysUnknownSchema).safeParse(input)
+      if (result instanceof Promise || ctxResult instanceof Promise) {
+        throw new TypeError('Schema validation must be synchronous')
+      }
```
gustavoguichard: We only allow sync validations at the moment but standard schema makes it easy to accept async.
Should we allow it?
Remember that the reasoning to remove it was that any async validation should go within the body of the function but allowing async validations will make reuse of schemas easier.

src/tests/constructors.test.ts
```diff
@@ -240,8 +242,8 @@ describe('withSchema', () => {
   })
 
   it('uses zod parsers to parse the input and context and call the schema function', async () => {
-    const parser = z.object({ id: z.preprocess(Number, z.number()) })
-    const ctxParser = z.object({ uid: z.preprocess(Number, z.number()) })
+    const parser = z.object({ id: z.coerce.number() })
+    const ctxParser = z.object({ uid: z.coerce.number() })
```
gustavoguichard: ~For some reason the standard-schema was not working with the preprocess.
We can discuss/investigate it further.~

Fixed

src/constructors.ts
diogob: I would allow to avoid the branch with the exception. I still believe it is a sort of anti-feature. But having this exception here hurts more

src/tests/constructors.test.ts
diogob: Besides this issue I don't see any breaking changes, am I missing something?

src/constructors.ts
```diff
@@ -1,10 +1,10 @@
 import { ContextError, ErrorList, InputError } from './errors.ts'
 import type { Internal } from './internal/types.ts'
+import type { StandardSchemaV1 } from '@standard-schema/spec'
```
diogob: What do you think of keeping the V1 only in the import? Looks a bit cleaner and reduces the delta in the case of upgrades.

```suggestion
import type { StandardSchemaV1 as StandardSchema } from '@standard-schema/spec'

```

src/tests/constructors.test.ts
gustavoguichard: @diogob the "breaking change" is that standard schema only works with zod@3.24+

gustavoguichard: Breaking changes:
- Zod at 3.24+
- ~preprocess seems to be broken~
- We don't export `ParserSchema` anymore

diogob (dismissed):
:tada:

PR #173, opened 2025-04-05
Fix async composable type

gustavoguichard:


gustavoguichard:
@diogob we should aim to satisfy this new test =)

src/tests/constructors.test.ts
```diff
@@ -81,6 +81,9 @@ describe('composable', () => {
     type _FN = Expect<
       Equal<typeof fn, Composable<(a: number, b: number) => number>>
     >
+    type _FN2 = Expect<
+      Unequal<typeof fn, Composable<(a: number, b: number) => Promise<number>>>
+    >
```
gustavoguichard, line 86: @diogob this is failing. If we can get this to pass we are gonna have fixed this weird bug.

PR #175, opened 2025-04-22
Performant `merge` and remove test duplication

gustavoguichard:
I'm just removing a duplicated test and adding a `merge` type written by [ahejlsberg himself](https://x.com/colinhacks/status/1912641412669272410).


diogob (commented):
Can't we replace all instances of Prettify?

